PR #67, opened 2025-01-20
chore: do not open Storybook automatically in the browser

raven-pan:
Storybook automatically opens default browser every time `pnpm ui-library:dev` is run.
After discussion, we have decided to disable this behavior.

reference: https://storybook.js.org/docs/api/cli-options


steven-chiu-bigstack (approved):
Say no more

https://github.com/user-attachments/assets/8142e028-1bf0-4eb6-a9e4-462c1177ffec

PR #69, opened 2025-01-21
feat: add the missing icons

lizzy-liang-bigstack:
- Update icons


steven-chiu-bigstack (approved):
The designers have renamed those icons. LGTM!👌

PR #72, opened 2025-01-22
feat: create CosCheckbox and its storybook

### LINKED ISSUE #15: UI Library - Checkbox
[Mockup](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=56-563&t=LfGu7yUKAaYr23yf-0)

lizzy-liang-bigstack:
## Summary:
- close #15 
- create `CosCheckbox` and `CosCheckboxGrid`

## Screenshots:

https://github.com/user-attachments/assets/092d1d31-2d22-47fe-b436-9a679f4b7a0a

## Usage:
- To display an indeterminate status, use the `indeterminate` prop.
- When `indeterminate={true}`, the following conditions will display the indeterminate icon:
   - Controlled input: `value = null`
   - Uncontrolled input: `defaultValue = null`
- When `indeterminate={false}` or the indeterminate flag is not set, it will display an unselected icon if `value = null`.

### DIFF
--- a/packages/cube-frontend-ui-library/src/components/CosCheckbox/CosCheckbox.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosCheckbox/CosCheckbox.tsx
@@ -1,5 +1,5 @@
-import { ChangeEvent, InputHTMLAttributes, useState, RefObject } from 'react'
 import { cva } from 'class-variance-authority'
+import { ChangeEvent, InputHTMLAttributes, RefObject, useState } from 'react'
 import { twMerge } from 'tailwind-merge'
 import CheckboxUnselected from '../../components/CosIcon/monochrome/checkbox.svg?react'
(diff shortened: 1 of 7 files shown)


steven-chiu-bigstack (commented):
Hi @lizzy-liang-bigstack 

Nice work!👍Just wanted to bring up something about the `indeterminate` prop.

It looks like we have covered 3 different states by defining the type of `checked` and `defaultChecked` props as `checked?: boolean | null`:

- `true`: Checked, shows the checked icon.
- `false`: Not checked, shows the empty box icon.
- `null`: Indeterminate, shows the half-checked icon.

With this setup, the `indeterminate` prop might not be necessary and can be removed. Does this make sense to you?

steven-chiu-bigstack (approved):
LGTM!

PR #93, opened 2025-02-14
feat(ui-library): 41 sidebar component

### LINKED ISSUE #41: UI Library - Sidebar
[Mockup](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=863-7191&m=dev)

raven-pan:
## SideBar - Group

image

## SideBar - Unit
Screen Recording 2025-02-14 at 1 30 05 PM

## Web APP Layout

Currently, I haven't integrated any real API data yet.
I will set that up in our next Typescript API SDK-related PRs.

image

closes #41


steven-chiu-bigstack (approved):
Looks great, nice work! 👍 
I just got some minor comments:

- I think `Combobox` is not a good name for the navlinks in the sidebar because [Combobox](https://support.microsoft.com/en-us/office/insert-a-combo-box-10c878be-9e6c-42c7-968f-76bc3df18c61) usually means a dropdown-like widget.
   - But that's how the designers call it in the mokcup. 🤷‍♂️ Let's leave it as is for now and only come back to rename it when necessary.
- I think it makes more sense to make `SideBarComboboxOption` always a link. We'll have to modify it when integrating React Router anyway, so this works for now. 👌 
- It's unfortunate that some text in the mockup are not aligned with the typography token (e.g., sidebar footer). But it's a minor problem, so just forget it! 😓

PR #95, opened 2025-02-17
feat: sidebar header api integration

raven-pan:
1. Generate a Typescript API SDK from [cube-cos-openapi](https://github.com/bigstack-oss/cube-cos-openapi) using [openapi-generator](https://github.com/OpenAPITools/openapi-generator). See [REAMD](https://github.com/bigstack-oss/cube-cos-ui/blob/e7c4f1b1042f5a0da468e811586283a6dc6e9129/packages/cube-frontend-api/README.md).
2. Automatically obtain an access token using `.env.local` (which is git-ignored) in local dev environment. See [README](https://github.com/bigstack-oss/cube-cos-ui/blob/e7c4f1b1042f5a0da468e811586283a6dc6e9129/packages/cube-frontend-web-app/README.md)
3. Integrate the Integrations, DataCenters and UserInfo API to properly display the header and sidebar. 

image


steven-chiu-bigstack (approved):
This is great, thanks for the effort!

I've pushed some commits to update the docs and code comments.
There are some minor questions regarding the SDK generation. I'll merge this PR first so that the development is not blocked.
We can deal with these questions later, don't worry about it! 🙏 

1. There's 1 line of rule in `packages/cube-frontend-api/sdk/.openapi-generator/FILES` that keeps getting removed every time I run `pnpm api:generate`. Does this happen on your end as well?
   image
2. I wonder how the files inside `packages/cube-frontend-api/sdk` were generated. It would be great if we can describe this (with 1 line) in `packages/cube-frontend-api/README.md`, and tell the developers not to remove any files within that folder.
3. Do you think it makes sense to add a `pnpm api:update` command that runs `git submodule update --recursive && pnpm api:generate`? This makes it slightly convenient to synchronize the status of the API when switching between branches (just a matter of personal preference, not a big deal 🤷‍♂️)

PR #103, opened 2025-02-18
Feat/cos api stream hook

raven-pan:
Please see the [COS API Request Hooks](https://github.com/bigstack-oss/cube-cos-ui/blob/a9f06f2d06f08922252616146ab8ca837b73aadc/packages/cube-frontend-web-app/docs/cos-api-request-hooks.md).

This hook will abort the request when the param change or the component is unmounted:

Screen Recording 2025-02-20 at 8 40 50 PM

Notes:
1. Currently, It does not support the error retry mechanism. According to the previous meeting, it should handle error retries and display a Top Nagging message to hint users that the UI data is not updated in real time and is attempting to reconnect.
2. The `fetchOnMount` and `fetchOnParamChanges` options are not actually used, so I have removed them for a simpler implementation.


steven-chiu-bigstack (approved):
Nicely done, it works great! 🚀  

Not sure if you’ve noticed this already, but your implementation does handle the "auto-refresh on/off" scenario:

https://github.com/user-attachments/assets/9d364ca3-f213-45b2-ba54-623ab0f6a170

The only small issue is that if we switch from "on" to "off" before the response arrives, it gets stuck in the loading state. A simple fix would be to disable the toggle while the stream request is still loading, which also makes sense from UI perspective.

There’s some room for minor refactoring (like de-duplicating code between different COS request hooks), but let’s not worry about that for now and keep focusing on feature development. 👌

PR #107, opened 2025-02-22
fix: cos stream request loading state incorrect

raven-pan:
When the unsubscribe occurs before the first loading is complete, the loading should be set to false.

Demo:

Screen Recording 2025-02-22 at 6 46 05 PM

reference: https://github.com/bigstack-oss/cube-cos-ui/pull/103#pullrequestreview-2631560568


steven-chiu-bigstack (approved):
LGTM!

PR #110, opened 2025-02-24
feat(ui-library): count segmented chart

raven-pan:
1. Implement the `CosCountSegmentedChart`, which is used on the Home and Chart pages.
2. Move `CosSegmented` to atom-level component.

image


steven-chiu-bigstack (approved):
LGTM! 👍

PR #111, opened 2025-02-24
feat(ui-library): percentage pie chart

raven-pan:
⚠️ This PR is based on https://github.com/bigstack-oss/cube-cos-ui/pull/110
---

Implement the `CosPercentagePieChart`, which is used on the Home and Chart pages.

image


steven-chiu-bigstack (approved):
Looks good, nice work!

I've pushed some commits to fix the minor issues mentioned in my other comments.

I remembered we had a discussion on the case where 99% would look like 100% due to our CSS implementation (98%. 99%. 100% looks almost the same on Windows by the way).

{F330E451-4684-4AC0-BDA9-EEBB46C16137}

I guess you removed the temporary fix for some reason, but since those are just edge cases it's totally fine to skip it and deal with it later. 👌

PR #112, opened 2025-02-24
feat(ui-library): 38 dashboard panel component

raven-pan:
⚠️ This PR is based on https://github.com/bigstack-oss/cube-cos-ui/pull/111

---

## Changes

Implement Dashboard Panel component.

## Demo

Screen Recording 2025-02-24 at 6 20 25 PM

## Known issues

See https://github.com/bigstack-oss/cube-cos-ui/issues/113.

PR #120, opened 2025-02-25
feat(ui-library): 29 UI library content switcher default

raven-pan:
⚠️ This PR is based on https://github.com/bigstack-oss/cube-cos-ui/pull/112

Demo
image


steven-chiu-bigstack (approved):
Good work! As per our discussion, the designers have updated the mockup after this PR is created last week.

I proceed to update the styles to align with the updated mockup. Let me know if you notice any problems! 🙏 

https://github.com/user-attachments/assets/a0fbb014-2f6e-4297-99d9-53c15680fdf5

PR #128, opened 2025-03-03
fix(ui-library): table show double border in panel

raven-pan:
fix(ui-library): table show double border in panel.

https://github.com/bigstack-oss/cube-cos-ui/pull/112#discussion_r1977134880

Screenshot:

image


steven-chiu-bigstack (approved):
Looks good, thanks for the quick fix!

PR #164, opened 2025-03-12
fix(web-app): cancel the stream reader when abort signal is issued & use single react version in a package

### LINKED ISSUE #160: React 19 Might Be Too New for Underlying Packages
---
name: Bug Report
about: Create a report to help us improve
title: '[Bug]: React 19 Might Be Too New for Underlying Packages'
labels: 'bug'
assignees: ''

---

## Describe the bug

I was trying to test and upgrade Keycloak version in `./packages/cube-frontend-keycloak-login`, and I encountered the error [Minified React error #525](https://react.dev/errors/525).

```
A React Element from an older version of React was rendered. This is not supported. It can happen if:
- Multiple copies of the "react" package is used.
- A library pre-bundled an old copy of "react" or "react/jsx-runtime".
- A compiler tries to "inline" JSX instead of using the runtime.
```

I tried checking out an older commit (99b04241a481032e0a516636e8b83444f652ade4), which made Keycloak login page working again. I have no idea what was going wrong, might need someone to look into, or I would come back to investigate what was going wrong after finished the urgent tasks.

## Version 

3.0.0 (dev) - cube-cos-ui (3e2bb79f5ba89d5b3f6807d4f9dba8432a1715c2).

## To Reproduce

1. `cd ./packages/cube-frontend-keycloak-login`
2. `pnpm run build && pnpm run copy-output && docker compose up`

## Expected behavior

Keycloak login page is shown.

Eandalf-Bigstack:
#### What type of PR is this?

- fix

- Cancel the stream reader when abort signal is issued
- Use single react version in a package

Fixes #160

### DIFF
--- a/packages/cube-frontend-web-app/src/hooks/useCosRequest/cosRequestUtils.ts
+++ b/packages/cube-frontend-web-app/src/hooks/useCosRequest/cosRequestUtils.ts
@@ -63,9 +63,13 @@ export const readStream = async <
   ChunkResponse extends CosApiInnerResponse<unknown>,
 >(
   stream: ReadableStream,
+  abortSignal: AbortSignal,
   onChunk: (chunkResponse: ChunkResponse) => void,
 ) => {
   const reader = stream.getReader()
+  abortSignal.onabort = () => {
+    reader.cancel()
+  }
   const decoder = new TextDecoder()
   let buffer = ''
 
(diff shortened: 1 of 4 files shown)


steven-chiu-bigstack (approved):
LGTM! 👍👍👍

Issue #132, opened 2025-03-04
UI Library - Table Empty State


raven-pan:
Implement the Table Empty State to enhance the user experience.

https://github.com/user-attachments/assets/a03fd68c-6d0c-4078-b6ea-b7eb5d1e0553

Issue #126, opened 2025-03-03
Page - Settings Page


steven-chiu-bigstack:
[Mockup](https://www.figma.com/design/kRJ5MUAkMF0AYgXCeChtZn/COS-management-UI-Redesign-v.3?node-id=2141-24654&p=f&t=F5Qe2mEExXogz176-0)

Dependencies:

- #43 
- #31 
- #4

PR #170, opened 2025-03-15
126 settings page slack channels

steven-chiu-bigstack:
This PR is for the Slack Channels section in the page.

The implementation is a little bit complicated because users can create, edit, try, and delete any row simultaneously. We can't reload the table after mutation as usual because that'll interrupt/erase users' inputs in the table.

## Skeletons

https://github.com/user-attachments/assets/07a96228-fd07-44a1-b6f2-90ebf7bf85ca

## Create (up to 10)

Users will have to re-enter the button with cursor to see the tooltip because `CosTooltip` utilizes `onMouseEnter` instead of `onMouseOver` (to increase performance). I guess this is okay for now? 🤷‍♂️ 

https://github.com/user-attachments/assets/bd3417b1-e180-44aa-a673-661eda89172c

## Try

https://github.com/user-attachments/assets/a94391e3-dceb-4dd8-8081-9ab3d0a7d16d

## Update

https://github.com/user-attachments/assets/b266a0ef-3470-4218-a4fe-553a1964d249

## Delete

The table layout shifts after a successful deletion, making it slightly inconvenient to delete multiple Slack channels. However, this use case should be rare, so I think we can leave it as is for now.

https://github.com/user-attachments/assets/265eaded-cda7-48e6-922e-2cad966fab26

## Known Issues

- #132


packages/cube-frontend-web-app/src/pages/settings/_components/validateBySchema.ts
```diff
@@ -0,0 +1,32 @@
...
+): ErrorRecord<T> => {
+  const errorRecord: ErrorRecord<T> = {}
+  const shape = schema.shape
+
+  for (const key in shape) {
+    const value = data[key]
+    const rule = shape[key]
+    const error = rule.safeParse(value).error?.issues[0]?.message
+    if (error) {
+      errorRecord[key as keyof T] = error
```
raven-pan: nitpick: we can safely remove the `as` cast here.

packages/cube-frontend-web-app/src/pages/settings/ManageContact.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const onTitlePrefixChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setTitlePrefix(e.target.value)
+  }
+
+  return (
+    <SettingsSection>
+      <h5 className="secondary-h4 text-functional-text">Manage Contact</h5>
+      <CosStroke type="dot" />
+      <div className="flex flex-col gap-y-1.5">
+        <div className="primary-body2 font-semibold text-functional-text">
```
raven-pan: We can use `<CosInput label="Title prefix" ... />` to assign input label.

raven-pan (approved):
LGTM! Just some minor nitpicks.

packages/cube-frontend-web-app/src/pages/settings/_components/validateBySchema.ts
steven-chiu-bigstack: You're right, let me remove it, thanks! 🙏

packages/cube-frontend-web-app/src/pages/settings/ManageContact.tsx
steven-chiu-bigstack: Right, I totally forgot about this. I'll use the `label` prop instead. 🙏

PR #171, opened 2025-03-16
feat(keycloak-login): build keycloak image with the custom theme and wrap it into an rpm

### LINKED ISSUE #84: CD - Wrap Build into RPM
## Wrap Build into RPM

- cube-cos-ui main part
- Keycloak theme

We currently have two separated projects, including the main part and the Keycloak theme for login pages. We need to wrap the builds into one or two RPMs and upload the RPMs to our local registry to be further downloaded and installed into CubeCOS during the build phase in the build jail. The requirement to be installed in the build jail applies specifically to the Keycloak theme part, since the registry to hold the Keycloak image only exists in the build jail.

We need to first decide our versioning strategy, either sharing the tag number across the two projects, or having their own version numbers in some places, like `package.json`. If we have the shared version number, then we can have one RPM. If we have two version numbers, then we could have two RPMs. Per discussion with @steven-chiu-bigstack , we are going toward the sharing version number strategy, right? Then, after confirming which parts of the build should be deployed, we could carry out the deploy job, like wrapping into an RPM, uploading to our local registry, and downloading and installing during CubeCOS build phase.

Eandalf-Bigstack:
#### What type of PR is this?

- feature

- Build Keycloak image with the custom theme and wrap it into an rpm
- Add documentations for building the rpm of the Keycloak login page and deploying a development build to a running CubeCOS
- Remove the redundant -e flag

Close #84


steven-chiu-bigstack (approved):
Feels good to me! 👍

PR #172, opened 2025-03-17
125 events page

### LINKED ISSUE #125: Page - Events Listing Page
[Mockup](https://www.figma.com/design/kRJ5MUAkMF0AYgXCeChtZn/COS-management-UI-Redesign-v.3?node-id=3657-12504&t=F5Qe2mEExXogz176-0)

Dependencies:

- #30 
- #17 
- #43 
- #25

lizzy-liang-bigstack:
## Change

- Create `Events` page and update query parameters based on the selected sub-tab
- Modify `CosDropdown`, `CosPagination` and `CosDatePicker` accordingly
- Add `zustand` package for global filter storage
- Close #125 

### Content Switcher

https://github.com/user-attachments/assets/0b52d296-94a8-4c5e-b2a8-9394a33577e1

### Table Filter

https://github.com/user-attachments/assets/57235bbd-b4f3-4c74-a868-0d8b12b70885

### Date Picker

https://github.com/user-attachments/assets/56f5a792-6ccd-4146-b8cb-407a251b93dd

### Pagination

⚠️ The `pagination` appears a bit off because the data from the API is not yet accurate. This issue has already been reported to the backend team.

## Scope

- `/events`


packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsTableFilter.tsx
```diff
@@ -0,0 +1,62 @@
...
+        isLoading={isEventsFilterLoading}
+        selectedStartDate={filters[eventsType]?.startDate}
+        selectedEndDate={filters[eventsType]?.endDate}
+        onChange={handleFilterChange}
+      />
+      <div className="h-[34px] border-l border-functional-border-divider"></div>
+      <Cancel
+        className="icon-md m-[10px] shrink-0 cursor-pointer text-functional-text-light"
+        onClick={handleFilterReset}
+      />
```
steven-chiu-bigstack, line 58: Even though this is not explicitly written in the mockup, I think it makes more sense to only show the X button when the filter is not empty.

packages/cube-frontend-ui-library/src/components/CosBasicTable/rendering/CosTableTdNoResult.tsx
steven-chiu-bigstack, line 1: It looks like this file is committed by mistake as the empty placeholder for tables is already handled in #178.

packages/cube-frontend-ui-library/src/components/CosDatePicker/CosDatePicker.tsx
```diff
@@ -87,6 +95,7 @@ export const CosDatePicker = (props: CosDatePickerProps) => {
       const isMenu = elementRef.current?.contains(target)
 
       if (!isTrigger && !isMenu) {
+        setIsClosedByOutsideClick(true)
```
steven-chiu-bigstack: Just curious, can't we just call `onOutsideClickClose?.()` from here? This place looks exactly where it belongs. 🤔 
It's okay to add `isCalendarOpen` and `onOutsideClickClose` to the dependencies of this `useCallback` because we need it:

```ts
const handleClickOutside = useCallback(
  (event: MouseEvent) => {
    const target = event.target as HTMLElement

    const isTrigger = anchorRef.current?.contains(target)
    const isMenu = elementRef.current?.contains(target)

    if (!isTrigger && !isMenu && isCalendarOpen) {
      setIsCalendarOpen(false)
      onOutsideClickClose?.()
    }
  },
  [anchorRef, elementRef, isCalendarOpen, onOutsideClickClose],
)
```

This way we can remove the new state above and the new effect below.

packages/cube-frontend-ui-library/src/components/CosPagination/utils.ts
```diff
@@ -5,6 +5,10 @@ type GetPageNumbersProps = {
 
 export type PageNumbers = 'ellipsis' | number
 
+export const DEFAULT_ITEMS_PER_PAGE_OPTIONS = [10, 20, 30, 50, 100]
+
+export const DEFAULT_ITEMS_PER_PAGE = 10
```
steven-chiu-bigstack: We do have these 2 constants in `packages/cube-frontend-ui-library/src/components/CosPagination/cosPaginationUtils.ts`:

{BC8A54B3-DB8F-4905-9DBC-6A1634987542}

We can try to use those ones as they're more type safe.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/utils.ts
```diff
@@ -0,0 +1,57 @@
+import dayjs from 'dayjs'
+
+const eventsRequestKeyMapping: Record<string, string> = {
```
steven-chiu-bigstack: Nitpick: We can use `Record<string, keyof EventsApiGetEventsRequest>` to make it more type safe, and to find problems more quickly if the API interface changes in the future.

`EventsApiGetEventsRequest` is from `@cube-frontend/api`.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsRefreshButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <div className="flex items-center gap-2">
+      {/** TODO: implement loading state to the button */}
+      {isEventsLoading && <CosLoadingSpinner variant="dot120" />}
+      <CosButton
+        type="secondary"
+        usage="icon-only"
+        Icon={ArrowRefresh}
+        onClick={handleEventsRefresh}
+      />
```
steven-chiu-bigstack: There's a `loading` prop on `CosButton` (I should've made the prop name more consistent by renaming it to `isLoading`. Sorry! 😬).
I didn't find related design in the mockup. Do you think it fits our use case?

chrome_p2LMjFBL4r

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsContentSwitcher.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { EventsFilterTableContext } from './context'
+
+export const EventsContentSwitcher = () => {
+  const { eventsType: activeTab, handleEventsTypeChange: handleTabChange } =
+    useContext(EventsFilterTableContext)
+  return (
+    <CosContentSwitcher variant="radius">
+      {Object.values(GetEventsTypeEnum).map((tab) => (
+        <CosContentSwitcher.Item
+          key={`tab-${tab}`}
```
steven-chiu-bigstack: Nitpick: I think `tab` alone is enough, but it doesn't really matter. 🤷‍♂️

packages/cube-frontend-utils/src/string.ts
```diff
@@ -3,3 +3,7 @@ import pluralize from 'pluralize'
 export const toUnitDisplay = (value: number, unit: string) => {
   return `${value} ${pluralize(unit, value)}`
 }
+export const toUpperCase = (str: string) => str.toUpperCase()
```
steven-chiu-bigstack: Nitpick: We can reuse the `toUpper` method from `lodash`.

packages/cube-frontend-utils/src/string.ts
```diff
@@ -3,3 +3,7 @@ import pluralize from 'pluralize'
...
+
+export const toUpperCaseFirstLetter = (str: string) =>
```
steven-chiu-bigstack: Nitpick: We can reuse the `capitalize` method from `lodash`.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsTableFilter.tsx
```diff
@@ -0,0 +1,62 @@
...
+    handleFilterReset,
+  } = useEventsFilter()
+
+  return (
+    <div className="flex items-center gap-3">
+      <div className="flex items-center gap-2">
+        <CosSearchBarFilter
+          isLoading={isEventsFilterLoading}
+          value={filters[eventsType]?.keyword || ''}
+          onChange={(e) => handleFilterChange({ keyword: e.target.value })}
```
raven-pan, line 29: We can use [debounce](https://lodash.com/docs/4.17.15#debounce) to improve the user experience and reduce API requests.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsTableFilter.tsx
```diff
@@ -0,0 +1,62 @@
...
+          onInputClear={() => handleFilterChange({ keyword: null })}
+          showDropdown={false}
+        />
+      </div>
+      {eventsFilter &&
+        eventsFilter[eventsType] &&
```
steven-chiu-bigstack: Nitpick: `eventsFilter && eventsFilter[eventsType]` can be shortened to `eventsFilter?.[eventsType]`

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/useEvents.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export const useEvents = (options: UseEventsOptions): UseEvents => {
+  const { eventsType, pageSize = 10, pageNum = 1 } = options
+
+  const { name: dataCenter } = useContext(DataCenterContext)
+
+  const { getFilters } = useEventsFilterStore()
+
+  const { data, isLoading, getResource } = useCosGetRequest(
```
steven-chiu-bigstack, line 55: (No need to deal with it in this PR)
We should add a cancellation mechanism to `useCosGetRequest` to ensure the promise callback runs only for the latest response. Otherwise we may observe behaviors like below:

https://github.com/user-attachments/assets/1f03a391-97d9-409a-b18a-fadf4b36a76e

I just created a new [issue](https://github.com/bigstack-oss/cube-cos-ui/issues/193) for it. 🙏

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/context.ts
steven-chiu-bigstack, line 1: I think it’s better to use either Context or Zustand within a single page, rather than mixing both, to keep things simpler. If we're going for Zustand, we should probably reset the filters when leaving the Events page.

The current implementation is a bit complicated. We can try to simplify the code by:

- Using query params to initialize the store on mounted.
- Then, use the store values as the source of truth for fetching events and synchronizing query string.

So all user inputs will be updated to the store, and we'll update the query string whenever the filter values of the current tab changes to avoid calling `setSearchParams` here and there.

---

One issue I've noticed is that the back and forward buttons didn't work after query params change:

https://github.com/user-attachments/assets/0f82bc6f-50f8-4d2c-9c8e-cf15052dd796

I haven't look into it yet. Let me know if you need any help! 🙏

packages/cube-frontend-web-app/src/stores/events.ts
```diff
@@ -0,0 +1,46 @@
+import { create } from 'zustand'
+
+type EventsFilterState = {
+  filters: Record<string, Record<string, string>>
+  setFilter: (tab: string, key: string, value: string) => void
+  clearFilter: (tab: string, key: string) => void
+  resetFilter: (tab: string) => void
+  getFilters: (tab: string) => Record<string, string>
+}
```
steven-chiu-bigstack: - It's recommended to change all the `tab: string` to aragument `tab: GetEventsTypeEnum` for type safety, including the key of the `filters` object.
- We can move the selected event type to the store. This way, methods that modify the filter (e.g., `setFilter`) can retrieve the `tab` argument internally, preventing accidental updates such as updating Host tab's filter while on the System tab.

For events chart page, using query string (to produce links connected with events page) would be more straightforward & easier to implement.

steven-chiu-bigstack (commented):
The UI looks good! I left some comments regarding the implementation detail. Let me know if you have any questions! 🙏

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsRefreshButton.tsx
lizzy-liang-bigstack: It does! Thank you for pointing it out 🥰

packages/cube-frontend-ui-library/src/components/CosBasicTable/rendering/CosTableTdNoResult.tsx
lizzy-liang-bigstack: Oops, my bad! Let me remove this one.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsTableFilter.tsx
lizzy-liang-bigstack: Sounds good! I think it will enhance the UX :)

packages/cube-frontend-ui-library/src/components/CosDatePicker/CosDatePicker.tsx
lizzy-liang-bigstack: There is a small problem with calling `onOutsideClickClose ` here. 
 
When the user presses the Apply or Cancel button, the `onOutsideClickClose ` function will also be invoked. Maybe we would expect it to be handled separately - I wonder if it makes sense 🤔

packages/cube-frontend-ui-library/src/components/CosPagination/usePageNumbers.ts
```diff
@@ -7,7 +7,7 @@ export type PageNumbers = 'ellipsis' | number
 
 const MAX_DISPLAY_PAGE = 7
 
-export const getPageNumbers = (props: GetPageNumbersProps): PageNumbers[] => {
+export const usePageNumbers = (props: GetPageNumbersProps): PageNumbers[] => {
```
steven-chiu-bigstack: Note: Usually, only hooks (built-in or custom) are prefixed with `use`. Custom hooks are just functions that use built-in or other custom hooks. Since this function doesn’t use any hooks internally, it feels more like a utility function rather than a hook (it can run anywhere, not just in React). But calling it `usePageNumbers` is fine. 👌

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsRefreshButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+type EventsRefreshButtonProps = {
+  onEventsMutate: (() => Promise<GetEventsResponseData>) | undefined
+  isEventsLoading: boolean
+}
+
+export const EventsRefreshButton = (props: EventsRefreshButtonProps) => {
+  const { onEventsMutate, isEventsLoading } = props
+
+  const handleEventsRefresh = () => {
+    onEventsMutate?.()
```
steven-chiu-bigstack: (Just a tiny issue) Sorry, I should've seen this in the last review:

Since clicking the button reloads the events table, names like `onRefreshClick` or `onReloadClick` might be more fitting.

The name `on[Resource]Mutate` sounds more like a callback that runs after a resource is mutated (created, updated, or deleted). Sometimes we reload a table or section after that, which is probably where the name came from.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsTableSelection.tsx
```diff
@@ -0,0 +1,118 @@
...
+  useEffect(() => {
+    setCurrentPage(pagination?.number ?? 1)
+  }, [pagination, setCurrentPage])
+
+  return (
+    <div className="flex flex-col gap-6">
+      <EventTable rows={rows} isLoading={isEventsLoading}>
+        <EventTable.Column
+          label={EventTableColumn.severity.label}
+          property={EventTableColumn.severity.property}
```
steven-chiu-bigstack, line 76: This works, but as the abstraction layers increase, readability may decrease. `property="severity"` is also type safe, and is more straightforward. That said, this is fine for now. 👌

packages/cube-frontend-web-app/src/stores/events.ts
```diff
@@ -0,0 +1,63 @@
...
+        ...state.eventsFilters,
+        [state.eventsType]: {},
+      }
+      return {
+        eventsFilters: updatedFilters,
+        isCurrentFilterEmpty: true,
+      }
+    }),
+  getEventsFilter: () => get().eventsFilters[get().eventsType] || {},
+  isCurrentFilterEmpty: true,
```
steven-chiu-bigstack: Similar to how you define `getEventsFilter`, `isCurrentFilterEmpty` should be a computed property based on the current filter. This way  we don't have to update both of them everywhere.

It would be great if there's a built-in computed property functionality in Zustand, but it doesn't look like it'll be added ([discussion](https://github.com/pmndrs/zustand/issues/132#issuecomment-678555466)).

So I think it would be better to change `isCurrentFilterEmpty` to a function and calls it in components:

```ts
isCurrentFilterEmpty: () => {
  const { eventsType, eventsFilters } = get()
  return isEmpty(eventsFilters[eventsType])
}
```

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/useEvents.ts
```diff
@@ -0,0 +1,60 @@
...
+        type: eventsType,
+        pageSize,
+        pageNum,
+      } satisfies EventsApiGetEventsRequest
+    },
+  )
+
+  return {
+    events: data?.events ?? [],
+    pagination: data?.page ?? ({} as Page),
```
steven-chiu-bigstack, line 74: Patterns like this will cause `pagination` to be a new object within every render if `data?.page` is `null` or `undefined`.

In `EventsTableSelection`, `pagination` is a dependency of an effect where we update a number state. This works for now, but we need to be careful since this pattern can easily cause infinite loops.

{E5E0FCC2-71D4-4CF0-96C2-3EF4F97F861C}

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/FilterDatePicker.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  const handleCancelClick = () => {
+    setStartDate(undefined)
+    setEndDate(undefined)
+
+    onDatePickerChange({
+      startDate: null,
+      endDate: null,
+    })
```
steven-chiu-bigstack, line 45: (Discussion, no need to make changes yet): When clicking cancel, do you think it makes sense to undo only the changes made in the current session rather than clearing the entire date picker? For me, cancel means...cancel 😅, while clear means clear.

packages/cube-frontend-ui-library/src/components/CosDatePicker/CosDatePicker.tsx
steven-chiu-bigstack: Hmm, as its' name suggested, the function is meant to be called when clicking outside of the datepicker. We've checked whether the clicked target is the trigger or the menu, so it should be safe to call `onOutsideClickClose` from here:

https://github.com/user-attachments/assets/bc512ff1-aca6-496d-b08c-d9826912c22f

packages/cube-frontend-web-app/src/stores/events.ts
```diff
@@ -0,0 +1,63 @@
...
+    [GetEventsTypeEnum.Instance]: {},
+  },
+  setEventsType: (type) => set({ eventsType: type }),
+  setEventsFilter: (key, value) =>
+    set((state) => {
+      const updatedFilters = {
+        ...state.eventsFilters,
+        [state.eventsType]: {
+          ...state.eventsFilters[state.eventsType],
+          [key]: value,
```
steven-chiu-bigstack: From the code it looks like `value` could be an empty string. In such case, we should delete the property from the `filter` object for the `isCurrentFilterEmpty` below to work as expected.

Alternatively, we can use the `clearEventsFilter` method below before updating the filter (in components) if the input value is empty, as the `clearEventsFilter` method is not used anywhere after the previous refactoring.

steven-chiu-bigstack (commented):
I think we can merge this PR after these two bugs are fixed. Other improvements can be made in separate issues (e.g., #195).

1. Search params are not retained after refresh

https://github.com/user-attachments/assets/6bac51b9-c199-4727-8efc-874df4cdf1ef

2. Back & forward buttons are not working in Events Page after params change; notice how I kept clicking on the back button but nothing happened.

https://github.com/user-attachments/assets/d87a3916-c5ba-4190-879f-a35ed9a22124

We can consider using the `replace` option to avoid pushing too many entries to the browser history:

{9A488731-D87A-4F00-A7F3-216A96EFE197}

---

By the way, [the docs](https://api.reactrouter.com/v7/types/react_router.SetURLSearchParams.html ) say we can use `setSearchParams` in the same way as `setState`—by passing an updater function that returns new search params. Maybe it'll come in handy sometimes:

{BD4996E2-99A1-47BF-8838-DCC90461B745}

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsRefreshButton.tsx
lizzy-liang-bigstack: You're right! The naming is changed as suggested.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/FilterDatePicker.tsx
lizzy-liang-bigstack: Perhaps I can open a ticket to track this discussion and confirm the user flow with the designers and team members in the next kickoff meeting?

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/useEvents.ts
lizzy-liang-bigstack: You're right! I've fixed it using `useMemo` + `useEffect` - I think this resolves the issue where the previous `useEffect` might have been excessively triggered when `data?.page` was `null` or `undefined`.

steven-chiu-bigstack: FYI: `useMemo` + `useEffect` is one way to do it, another solution is to change the type of `pagination` from `Page` to `Page | undefined` and do the null check in the effect callback. This means we can remove the `({} as Page)` default value and let it be `undefined` before the API responds:

```ts
useEffect(() => {
  const currentPage = pagination?.number ?? 1
  setCurrentPag(pagination)
}, [pagination])
```

For places that's also accessing `pagination`, the can use optional chaining and null coalesce to prevent errors. For example, `pagination?.number ?? 1`, `page?.totalItemCount ?? 0`, etc.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/FilterDatePicker.tsx
steven-chiu-bigstack: Sure, no problem! 👍

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsTableFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  eventsType: GetEventsTypeEnum
+}
+
+export const EventsTableFilter = (props: EventsTableFilterProps) => {
+  const { eventsType } = props
+
+  const { isLoading: isEventsFilterLoading, getEventsFilter } =
+    useEventsFilter()
+
+  const { getCurrentQuery } = useEventsQuery()
```
steven-chiu-bigstack: Nitpick: Currently `useEventsQuery().getCurrentQuery()` is called in both `EventsTableFilter` and `useEvents`. This is fine for now as it's not a heavy computation, but we can avoid duplicating the computation by putting it in `useEvents` and pass it down to `EventsTableFilter`.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsTableSelection.tsx
```diff
@@ -0,0 +1,127 @@
...
+   * if data?.page is null or undefined.
+   */
+  const stablePagination = useMemo(
+    () => pagination ?? { number: 1, totalItemCount: 0 },
+    [pagination],
+  )
+
+  useEffect(() => {
+    setCurrentPage(stablePagination.number)
+  }, [stablePagination, setCurrentPage])
```
steven-chiu-bigstack: I wonder what this effect and `stablePagination` do? 🤔
Shouldn't the frontend `currentPage` state control the active page of the pagination component instead of using the `number` from the API response?

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/EventsTableSelection.tsx
```diff
@@ -0,0 +1,127 @@
...
+  currentPage: number
+  setCurrentPage: React.Dispatch<React.SetStateAction<number>>
+  itemsPerPage: ItemsPerPage
+  setItemsPerPage: React.Dispatch<React.SetStateAction<ItemsPerPage>>
+  events: GetEventsResponseDataEventsInner[] | undefined
+  pagination: Page | undefined
+  isEventsLoading: boolean
+}
+
+export const EventsTableSelection = (props: EventsTableSelectionProps) => {
```
steven-chiu-bigstack, line 52: Just curious, did you mean "Section" instead of "Selection"?

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/useEventsQuery.ts
```diff
@@ -0,0 +1,95 @@
...
+      } else {
+        newParams.delete(key)
+      }
+    })
+    setSearchParams(newParams, { replace: true })
+  }
+
+  const handleEventsQueryReset = () => {
+    const newParams = new URLSearchParams()
+    newParams.set('eventType', eventsType)
```
steven-chiu-bigstack: I noticed a bug where resetting the filter also resets the selected tab:

https://github.com/user-attachments/assets/8f5626e6-8f88-4d05-9d90-b2e8f4b6c404

This happens because the param name is inconsistent (`eventsType` and `eventType`.) We can define a constant to improve this.

---

By the way, I think we should reset the `currentPage` state to 1 after filter value changes (filter changes -> set current page to 1 -> then call the API):

https://github.com/user-attachments/assets/dc208d63-89f0-42e5-a46d-27069bcc0dd6

---

By the way again, there's an error (due to response with 400 status code) that says "pageNum should be provided if pageSize is provided" when entering the Events page:

https://github.com/user-attachments/assets/3963554a-9c2a-4be8-b635-838b016f86f8

steven-chiu-bigstack (approved):
Hi @lizzy-liang-bigstack 

Nice work, very close to done!
I noticed some new bugs (mentioned in [one of](https://github.com/bigstack-oss/cube-cos-ui/pull/172#discussion_r2020117861) the new comments). Feel free to merge this PR once they're all fixed, thank you for the hard work! 🙏

PR #173, opened 2025-03-17
126 settings page email recipients

steven-chiu-bigstack:
# ⚠️This PR is based on #191⚠️

This PR is for the Slack Channels section in the page.

The implementation is a little bit complicated because users can create, edit, try, and delete any row simultaneously. We can't reload the table after mutation as usual because that'll interrupt/erase users' inputs in the table.

## Sleletons

https://github.com/user-attachments/assets/e2108846-f17e-4143-93d8-c53148ef877d

## Create (up to 10)

https://github.com/user-attachments/assets/c34821f0-5afa-4527-a1b9-e681ff00e50d

## Try

https://github.com/user-attachments/assets/bb787418-d605-43c4-9dd4-90bd0165b79f

## Update

https://github.com/user-attachments/assets/2ab16a8d-2c80-4c43-aa34-5d837934a266

## Delete

https://github.com/user-attachments/assets/d910aea0-a98d-498f-a2dc-d9aa29189415

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/settings/_components/email/EmailSettings.tsx
+++ b/packages/cube-frontend-web-app/src/pages/settings/_components/email/EmailSettings.tsx
@@ -23,7 +23,11 @@ export const EmailSettings = (props: EmailSettingsProps) => {
 
   return (
     <div className="flex flex-col gap-y-6">
-      <EmailRecipients hasVerifiedSender={hasVerifiedSender} />
+      <EmailRecipients
+        isLoading={isLoading}
+        initialRecipients={initialData?.recipients}
+        hasVerifiedSender={hasVerifiedSender}
+      />
       <EmailSenders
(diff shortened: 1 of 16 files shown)


raven-pan (approved):
Nice work!

There's no comment on the code, but I noticed that the new email recipients appear as the last row after page is refreshed:

https://github.com/user-attachments/assets/8d2dcb92-1444-4bf5-9b63-3bfb577b6548

it's not a bug, but we could improve this user experience if we have time.

steven-chiu-bigstack:
> I noticed that the new email recipients appear as the last row after page is refreshed:

You're right, that's because the API sorts email recipients by `name` in ascending order.

I've thought about sorting the rows with the same rule after insertion during implementation, but that'll change the row order while the user is editing, which feels a bit off. Let's stick to the current implementation and see how things goes. 🙏

Issue #22, opened 2025-01-05
UI Library - Notification Inline


steven-chiu-bigstack:
[Mockup](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=501-819&t=LfGu7yUKAaYr23yf-0)

PR #184, opened 2025-03-18
127 tunings page

### LINKED ISSUE #127: Page - Tunings Page
[Mockup](https://www.figma.com/design/kRJ5MUAkMF0AYgXCeChtZn/COS-management-UI-Redesign-v.3?node-id=3679-26264&p=f&t=F5Qe2mEExXogz176-0)

Dependencies:

- #25

steven-chiu-bigstack:
Close #127 

https://github.com/user-attachments/assets/55961080-c77b-4284-a82b-014df24a3f07

## Skeleton

https://github.com/user-attachments/assets/52b67556-0847-44d3-99ca-81272fff36fe

## Known Issues

- #193 
- Keyword search does not always return the expected results at the moment due to the API implementation. This will be handled in [another issue](https://github.com/bigstack-oss/cube-cos-api/issues/154).
- The `status.isUpdating` field does not always work as expected in the API.


packages/cube-frontend-web-app/src/pages/events/tunings/TuningsPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+      <CosModal
+        title="Reset Tuning"
+        size="sm"
+        isOpen={isResetModalOpen}
+        actionText="Reset"
+        onActionClick={onConfirmReset}
+        onCloseClick={onCloseResetModal}
+      >
+        Are you sure you want to reset this tuning to default?
+      </CosModal>
```
lizzy-liang-bigstack: I think we need to add styles to the content text here. 
According to the design, it seems to be: `primary-body4 text-functional-text`.

- Current UI:
Screenshot 2025-04-09 at 4 39 22 PM

- Design
Screenshot 2025-04-09 at 4 42 49 PM

packages/cube-frontend-ui-library/src/components/CosModal/CosModal.tsx
```diff
@@ -13,12 +13,16 @@ export type CosModalProps = {
...
+  isActionButtonVisible?: boolean
   /**
    * @default 'Action'
    */
   actionText?: string
   actionButtonProps?: Pick<CosButtonProps, 'loading' | 'disabled'>
-  onActionClick: () => void
+  onActionClick?: () => void
   onCloseClick: () => void
 }
```
lizzy-liang-bigstack, line 27: Do you think it makes sense - to pass `onActionClick` only when `isActionButtonVisible` is `true`?

lizzy-liang-bigstack:
Just wanted to check—will the failed message implementation be handled in a separate ticket after the Notification components are completed?

Screenshot 2025-04-09 at 5 30 12 PM

packages/cube-frontend-web-app/src/pages/events/tunings/TuningsPage.tsx
steven-chiu-bigstack: Good catch! I'll update the styles accordingly. 🙏

packages/cube-frontend-ui-library/src/components/CosModal/CosModal.tsx
steven-chiu-bigstack: Yes, but that'll make it less convenient if we ever want to show/hide the action button dynamically (we don't have that case yet.) For example, the following code should fail the TS check if we make the props typing stricter:

```tsx
const isActionButtonVisible = useMemo<boolean>(() => {
  return someBooleanCalculatedBasedOnOtherPropsOrStates
}, [...])

const onActionClick = () => {
  // ...
}

<CosModal isActionButtonVisible={isActionButtonVisible} onActionClick={onActionClick} />
```

The current interface makes TS accept the above code, so I'd prefer to keep it, if you don’t mind.

steven-chiu-bigstack:
> Just wanted to check—will the failed message implementation be handled in a separate ticket after the Notification components are completed?

Right, I totally forgot about this! 💫 I think the designers added it to the mockup after this page is implemented. I'll update this PR later today to fix this (using some simple `div`s as notification before #22 is completed.)
Thanks for the reminder! 🙇‍♂️

packages/cube-frontend-ui-library/src/components/CosModal/CosModal.tsx
lizzy-liang-bigstack: No problem! 💃

packages/cube-frontend-web-app/src/pages/events/EventsLayout.tsx
```diff
@@ -13,7 +13,7 @@ export const EventsLayout = () => {
   const location = useLocation()
 
   return (
-    <div>
+    <div className="flex flex-col gap-y-4 px-2 py-1">
```
raven-pan, line 16: We need to remove the margin top from Event Chart page, so they can share same row gap.

https://github.com/user-attachments/assets/8b80b05e-4b2b-470f-908f-c9adb6c4ca32

packages/cube-frontend-web-app/src/pages/events/tunings/useTuningRows.ts
```diff
@@ -0,0 +1,149 @@
...
+        status: {
+          ...row.status,
+          isUpdating: false,
+        },
+      })
+    }
+  }
+
+  return {
+    isLoading: !hasResponseBeenReceived,
```
raven-pan, line 165: This will cause the tuning table to not show the loading state after the user changes the filter condition or pagination.
However, we can’t use `isLoading` because we have 5-second polling in the UI.
Maybe we can open a separate ticket to handle this issue later?


https://github.com/user-attachments/assets/29310f1a-be91-4491-a260-9793fdef32af

packages/cube-frontend-web-app/src/pages/events/tunings/_components/tableCells/ActionCell.tsx
```diff
@@ -0,0 +1,98 @@
+import {
+  CosToggle,
+  CosTooltip,
+  CosTooltipInformation,
+} from '@cube-frontend/ui-library'
+import ArrowRotateLeft from '@cube-frontend/ui-library/icons/monochrome/arrow_rotate_left_01.svg?react'
+import Edit from '@cube-frontend/ui-library/icons/monochrome/edit.svg?react'
+import { IconActionButton } from '@cube-frontend/web-app/pages/settings/_components/IconActionButton'
```
raven-pan: we can move the file to common components folder.

raven-pan (approved):

packages/cube-frontend-web-app/src/pages/events/EventsLayout.tsx
steven-chiu-bigstack: No problem, I'll remove it accordingly, thanks! 🙏

packages/cube-frontend-web-app/src/pages/events/tunings/useTuningRows.ts
steven-chiu-bigstack: Thanks for bringing this up! It looks like there are 2 kinds of loading:

1. First time loading (no response has ever been received): Skeletons are okay here.
2. Any subsequent loading: Skeletons are not ideal, especially for polling with short interval.

We can discuss with the designers what should we do in case #2. Some common solutions are:

1. Makes the table slightly transparent.
2. Covers the table with a mask that has a spinner in the center.

I've created an [issue](https://github.com/bigstack-oss/cube-cos-ui/issues/237) for it.

packages/cube-frontend-web-app/src/pages/events/tunings/_components/tableCells/ActionCell.tsx
steven-chiu-bigstack: No problem. 👌

steven-chiu-bigstack:
Hi all.

I've made some changes based on your feedback. The inline error notifications are also added to the Tunings page (those messages are parsed directly from the API response; I didn't capitalized them):

https://github.com/user-attachments/assets/4dd07aab-933e-43ab-969f-d059da99cd80

This PR is ready for review again, thanks! 🙏

PR #185, opened 2025-03-18
33 create CosTextArea

### LINKED ISSUE #33: UI Library - Textarea
[Mockup](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=54-844&t=LfGu7yUKAaYr23yf-0)

lizzy-liang-bigstack:
## Changes
- Create `CosTextArea`
- Close #33 

### Master

https://github.com/user-attachments/assets/30681e2b-edb4-4cda-bd97-f562a02045e4

### Status

https://github.com/user-attachments/assets/398ec045-f876-4e71-9f56-cd2fbf318b33

### Skeleton

Screenshot 2025-03-18 at 6 06 59 PM


packages/cube-frontend-ui-library/src/components/CosTextArea/CosTextArea.tsx
```diff
@@ -0,0 +1,117 @@
...
+      {/**
+       * Mirror div is set for calculation of the displayed texts
+       * When textarea is not focused, displays text according to the textarea  size
+       */}
+      <div
+        ref={mirrorRef}
+        className="pointer-events-none absolute left-0 top-0 w-full whitespace-pre-wrap break-words opacity-0"
+      ></div>
+      <textarea
+        {...props}
```
steven-chiu-bigstack: This will apply all props (including the non-builtin ones) to the `textarea`, causing React to log errors to the console:

{48DA4297-D638-48B3-8C7B-0859A546EE52}

So we should still use the `restProps` from `const { a, b, c, ...restProps } = props` and pick necessary props manually.

packages/cube-frontend-ui-library/src/components/CosTextArea/useTextAreaOverflow.ts
```diff
@@ -0,0 +1,57 @@
+import { useEffect, useRef, useState } from 'react'
+import { normalizeValue } from './cosTextAreaUtils'
+
+interface UseTextAreaOverflowProps {
+  valueProps: string | number | readonly string[] | undefined
+  maxLength: number
+}
+
+export const useTextAreaOverflow = (props: UseTextAreaOverflowProps) => {
```
steven-chiu-bigstack: To make the current solution more reliable, except for `width`, we should also make sure the `padding`, `font-family`, `font-size`, and `line-height` are identical between the mirror div and the textarea.

The current implementation has a few downsides:

- When blurred, the textarea acts as if it’s scrolled to the top:
  - chrome_6WK5AkoV1T
- The tuncated text should've been re-computed when the textarea is resized when it's not focused:
  - chrome_vYH3jJhKXz
- Performance doesn’t scale well as the max length increases. The video below shows a noticeable freeze when the textarea is blurred, caused by a while loop that trims/truncates the text from the end:

https://github.com/user-attachments/assets/72a1414c-5c92-42e9-8f6e-293016ba76eb

---

A more performant & easier solution (I hope so 😓) is to calculate how many rows are visible based on the computed CSS values, and use `line-clamp` to append the trailing ellipsis at the end of multi-lines text (it's still not perfect; need some polishing).

chrome_vQ7nhQfTw6

The height of each row/line in `textarea` is determined by `line-height`. We can get the computed value via [getComputedStyle](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle), and use the height of the textarea to calculate how many lines are visible.

The following code is just an example (half pseudo-code). You can refactor it to make it more maintainable. Feel free to let me know if you have any questions! 🙏 

```tsx
const CosTextArea = (props: CosTextAreaProps) => {
  const { ref, value } = props
  
  const textareaRef = useRef<HTMLTextAreaElement | null>(null)
  const [visibleRowsCount, setVisibleRowsCount] = useState(0)

  useEffect(() => {
    const textarea = textareaRef.current
    let observer: ResizeObserver | undefined = undefined

    if (textarea) {
      observer = new ResizeObserver((entries) => {
        const { lineHeight } = getComputedStyle(textarea)
        const lineHeightPx = parseInt(lineHeight)
        const textareaHeight = entries[0]?.contentBoxSize[0]?.blockSize ?? 0
        setVisibleRowsCount(
          Math.ceil((textareaHeight + textarea.scrollTop) / lineHeightPx),
        )
      })

      observer.observe(textarea)
    }

    return () => {
      observer?.disconnect()
    }
  }, [])

  return (
    <textarea
      ref={(element) => {
        // Please see `packages/cube-frontend-ui-library/src/components/CosTooltip/CosTooltip.tsx:89`.
      })
      value={value}
      style={isFocused ? undefined : LINE_CLAMP_CSS_PROPERTIES}
    />
  )
}
```

For `LINE_CLAMP_CSS_PROPERTIES`, please see [here](https://v3.tailwindcss.com/docs/line-clamp). Tailwind is also using these Webkit CSS properties, so I think we can ignore those deprecated warning (there's currently no alternatives 😕).

packages/cube-frontend-ui-library/src/components/CosTextArea/CosTextArea.tsx
```diff
@@ -0,0 +1,117 @@
+import { ChangeEvent, useState } from 'react'
+import { cva } from 'class-variance-authority'
+import { twMerge } from 'tailwind-merge'
+import { CosTextAreaSkeleton } from './CosTextAreaSkeleton'
+import { useTextAreaOverflow } from './useTextAreaOverflow'
+
+const textarea = cva(
+  [
+    'primary-body2 rounded-[5px] bg-grey-0 px-4 py-[10px] outline-none',
+    'max-h-[260px] min-h-[140px] min-w-[280px]',
```
steven-chiu-bigstack, line 11: Discussion: There's a `max-width: 380px` requirement for textarea in the mockup, but the corresponding class is not added here.

{A37AD76C-4E6A-4159-B10B-064CA5B40CB5}

Is it correct that we have a similar case in `CosTableInput`? If we were to restrict the max-width from the parent, does that mean users will have to add `max-w-[380px]` to the parent of every `CosTextArea`? If it is, do you think it make sense to add `w-full max-w-[380px]` to `CosTextArea` instead?

steven-chiu-bigstack (approved):
Hi @lizzy-liang-bigstack 

Thanks for the great work! I left some comments for you to check. Feel free to let me know if you have any questions! 🙏

packages/cube-frontend-ui-library/src/components/CosTextArea/CosTextArea.tsx
lizzy-liang-bigstack: Ah, you're right! I'll update its style.
 
However, I didn’t add a max width to it because it’s not applicable in known use cases.
  
For example, on the `Create Event Triggers` page, the width seems to exceed `380px` 🤔
 
Screenshot 2025-03-20 at 2 55 08 PM

steven-chiu-bigstack: Got it, that's unfortunate. 😕
Let's forget about `max-width` for now.

steven-chiu-bigstack (approved):
Looks good! I did some minor refactoring ([commit](https://github.com/bigstack-oss/cube-cos-ui/pull/185/commits/249e14701454c499c342ffb8041becfdf68d9f9f)), going to merge this now. 🙏

PR #186, opened 2025-03-18
Feat/124 chart page storage and ranking panels

### LINKED ISSUE #124: Page - Chart Page
[Mockup](https://www.figma.com/design/kRJ5MUAkMF0AYgXCeChtZn/COS-management-UI-Redesign-v.3?node-id=3562-28804&p=f&t=F5Qe2mEExXogz176-0)

Dependencies:

- #110 
- #111

raven-pan:
### UI Library - General Panel support Hyperlink

Screenshot 2025-03-19 at 2 45 08 PM

### Chart Page Storage and Ranking Panels

https://github.com/user-attachments/assets/35b774a9-812c-42ba-a13a-552fd15389f1

## Notes

1.  I chose [Chart.js](https://www.chartjs.org/) and [react-chartjs-2](https://react-chartjs-2.js.org/) to implement the [Line Chart](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=6359-15373&m=dev). Originally, the design guidelines were not created specifically for the `Chart.js` library. So there will be some minor differences between the actual appearance and the design guidelines. I will do my best to make them looks as similar as possible.
2. Due to the limitations of `Chart.js`, tooltips cannot displayed outside the canvas. Therefore, I had to minimize the Ranking Chart tooltip to fit within the canvas.

## TODOs, Discussion and Known Issues

All TODOs, discussion and known issues will be documented in this ticket: https://github.com/bigstack-oss/cube-cos-ui/issues/190 

Closes #124


packages/cube-frontend-web-app/src/hooks/useInterval.ts
```diff
@@ -0,0 +1,10 @@
+import { useEffect } from 'react'
+
+export const useInterval = (callback: () => void, delay: number) => {
+  useEffect(() => {
+    const interval = setInterval(() => {
+      callback()
+    }, delay)
+    return () => clearInterval(interval)
+  })
```
steven-chiu-bigstack: It looks like the interval is going to be reset after every render (even if `callback` and `delay` do not change). It's probably safer to only reset the interval when `callback` or `delay` changes (we can avoid putting `callback` in the dependencies with `useSyncedRef` from `packages/cube-frontend-utils/src/reactRefUtils.ts`):

```ts
export const useInterval = (callback: () => void ,delay: number) => {
  const callbackRef = useSyncedRef(callback)

  useEffect(() => {
    const interval = setInterval(() => {
      callbackRef.current()
    }, delay)
    return () => clearInterval(interval)
  }, [delay])
}
```

By the way, normally we would call `callback()` before/after setting up the interval so it runs immediately the first time. However, `useCosGetRequest` already done that for us.

If we want to make this hook more flexible, we could add an option to control whether the callback should be run immediately when setting up the interval.

packages/cube-frontend-web-app/src/pages/home/chart/_components/StoragePanels/useMetricsParams.ts
```diff
@@ -0,0 +1,33 @@
...
+import dayjs from 'dayjs'
+import { MetricsApiGetMetricByTypesRequest } from '@cube-frontend/api'
+import { DataCenterContext } from '@cube-frontend/web-app/context/DataCenterContext'
+import { GetParamFn } from '@cube-frontend/web-app/hooks/useCosRequest/cosRequestUtils'
+import { TypeParams } from '../utils'
+
+export const useMetricsParams = () => {
+  const dataCenter = useContext(DataCenterContext)
+
+  const currentTime = useRef(dayjs())
```
steven-chiu-bigstack, line 11: (Just FYI, no need to make changes because it makes the code less readible)

I've learned that we can use the lazy initialization pattern to avoid the initialization function of `useRef` from being called in every render:

```ts
const currentTime = useRef<Dayjs | undefined>(undefined)

const getCurrentTime = (): Dayjs => {
  if (!currentTime.current) {
    currentTime.current = dayjs()
  }
  return currentTime.current
}
```

Then we only use `getCurrentTime()` to access `currentTime.current`.

packages/cube-frontend-web-app/src/pages/home/chart/_components/StoragePanels/useMetricsParams.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const getTimeRangParams = () => {
```
steven-chiu-bigstack: Typo: `Rang` -> `Range`.

packages/cube-frontend-web-app/src/pages/home/chart/_components/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+  GetNetworkTrafficOutRankOfVmsResponse,
+  GetCpuUsageRankOfVmsResponse,
+  GetMemoryUsageRankOfVmsResponse,
+  GetDiskReadIopsRankOfVmsResponse,
+  GetDiskWriteIopsRankOfVmsResponse,
+} from '@cube-frontend/api'
+import { metricsApi } from '@cube-frontend/web-app/api/cosApi'
+import { CosApiResponse } from '@cube-frontend/web-app/hooks/useCosRequest/cosRequestUtils'
+
+export const getMetricsByTypes = async <T extends GetMetricByTypes200Response>(
```
steven-chiu-bigstack: Nitpick: We can remove `export` to keep this helper function internal.

packages/cube-frontend-web-app/src/pages/home/chart/_components/RankingPanels/RankingChart/RankingChart.tsx
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  return (
+    <div className="flex flex-col gap-y-4">
+      <div className="flex flex-col gap-y-3">
+        {isLoading
+          ? range(5).map((i) => <RankingItemSkeleton key={i} />)
+          : ranking.rank.map((rankingItem) => (
+              <RankingItem
+                key={rankingItem.name}
```
steven-chiu-bigstack: I saw the description you wrote in #190. If you're planning to merge this PR before receiving responses from the backend developers, it's suggested that we find a way to compose unique keys for each `RankingItem` to ensure the items rendered on the page is correct.

steven-chiu-bigstack (approved):
Looks good, thanks for the hard work! 👍👍👍 
No major concern from me except for the duplicate keys on `RankingItem`, as it causes the rendered items to be incorrect. Feel free to merge this PR after that one is fixed, either temporarily or permanently. 🙏

packages/cube-frontend-web-app/src/pages/home/chart/_components/StoragePanels/useMetricsParams.ts
raven-pan: Thanks for sharing!
I'll use the lazy initialization pattern in scenarios with expensive initialization in the future.

packages/cube-frontend-web-app/src/pages/home/chart/_components/RankingPanels/RankingChart/RankingChart.tsx
raven-pan: Further discussion here: https://github.com/bigstack-oss/cube-cos-ui/issues/190#issuecomment-2739935125

PR #191, opened 2025-03-20
126 settings page email senders

steven-chiu-bigstack:
# ⚠️This PR is based on #170⚠️

This PR is for the Email Senders section in the page.

## Skeletons

https://github.com/user-attachments/assets/420ae1f8-7630-426c-86ba-e964f131e0c3

## Create from Placeholder Row

https://github.com/user-attachments/assets/55c8e1a8-a6e3-42b4-8358-de2054d93830

## Update

I haven't discussed with the designers about the layout of editing an existing row. We can do that later.

https://github.com/user-attachments/assets/c6b5eeb4-c329-4b11-93eb-4539b79400de

## Verify - Success (mock)

https://github.com/user-attachments/assets/8b894f5d-2eb7-4e75-8a39-46392534f48e

## Verify - Fail

https://github.com/user-attachments/assets/6e322fb4-ef79-42f3-b87b-ff6bf5560718


packages/cube-frontend-ui-library/src/components/CosTableInput/CosTableInput.tsx
```diff
@@ -50,7 +50,7 @@ export const CosTableInput = forwardRef<HTMLInputElement, CosTableInputProps>(
 
     const renderErrorIcon = () => {
       return (
-        <div className="absolute right-0 flex size-4 shrink-0 translate-x-6 items-center justify-center overflow-hidden [&>*]:size-4">
+        <div className="flex size-4 shrink-0 items-center justify-center">
```
steven-chiu-bigstack, line 53: I changed the error icon from absolute position to flex so that it doesn't block the content to the right of the input.

Before:

costableinput_before

After:

costableinput_after

packages/cube-frontend-web-app/src/pages/settings/_components/email/senders/VerifyEmailSenderModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  return (
+    <CosModal
+      title="Send Test Email"
+      size="sm"
+      isOpen={isOpen}
+      actionText="Send"
+      actionButtonProps={{
+        loading: isTrying,
+        disabled: !sendTo || !!inputError,
```
raven-pan: nitpick: since we already validate the `sendTo` value, maybe we can just use `disabled: !!inputError,` to determine the disabled state.

packages/cube-frontend-web-app/src/pages/settings/_components/email/senders/VerifyEmailSenderModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  const onChange = (e: ChangeEvent<HTMLInputElement>): void => {
+    const { value } = e.target
+    setSendTo(value)
+  }
+
+  const onKeyDown = (e: KeyboardEvent<HTMLInputElement>): void => {
+    if (e.key === 'Enter' && sendTo && !inputError) {
```
raven-pan: nitpick: since we already validate the `sendTo` value, maybe we can just use `if (e.key === 'Enter' && !inputError)` to determine whether the user can submit the form.

raven-pan (approved):
Looks good overall!

Just a few nitpicks, and I noticed some weird behavior:

https://github.com/user-attachments/assets/37ade369-ca15-4d7f-81f1-f5137825db30

It's okay to handle that in another PR. Thanks!

packages/cube-frontend-web-app/src/pages/settings/_components/email/senders/VerifyEmailSenderModal.tsx
steven-chiu-bigstack: You're right, I'll remove those unnecessary checks. 🙏

steven-chiu-bigstack:
Thanks for pointing out the bug! 👍
It happened because of the `...row` statement in the `updateEmailSender` action, which has been fixed just now.

PR #203, opened 2025-03-27
refactor(keycloak-login): use topdir macro & bump the version to force update images on helm

Eandalf-Bigstack:
#### What type of PR is this?

- refactor

- Use `_topdir` macro to allow using a local rpm build directory in CubeCOS build process
- Bump the version to force update images on helm


steven-chiu-bigstack (approved):
LGTM! 👍👍👍

PR #205, opened 2025-03-27
CI - Create RPMs for UI and LOGIN Using Workflows

Eandalf-Bigstack:
#### What type of PR is this?

- ci

- Create rpms for UI and LOGIN using workflows

### DIFF
--- a/.github/workflows/cube-cos-login-rpm.yaml
+++ b/.github/workflows/cube-cos-login-rpm.yaml
@@ -87,7 +87,8 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           BUILD=$(ls ~/rpmbuild/RPMS/x86_64/ | cut -d" " -f1)
+          mv "~/rpmbuild/RPMS/x86_64/$BUILD" ./
           gh release create "cube-cos-login-$VERSION-dev-$BUILD_NUMBER" \
             --repo="$GITHUB_REPOSITORY" \
             --generate-notes \
-            "~/rpmbuild/RPMS/x86_64/$BUILD"
+            "./$BUILD"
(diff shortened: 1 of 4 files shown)


.github/workflows/cube-cos-login-rpm.yaml
```diff
@@ -0,0 +1,92 @@
...
+          cd rpmbuild
+          mkdir BUILD RPMS SOURCES SPECS SRPMS
+          popd
+
+      - name: Prepare Source Tarball
+        run: |
+          mkdir ~/source
+          mkdir -p ~/source/packages/cube-frontend-ui-library
+          pushd ./packages/cube-frontend-ui-library
+          cp -r ./src package.json postcss.config.js tailwind.config.js tsconfig.json vite.config.ts ~/source/packages/cube-frontend-ui-library
```
steven-chiu-bigstack, line 50: Discussion: Just curious, do you think there's enough time to define some ignore patterns for these copy commands so we don’t have to pick each file one by one? For example, reusing the grep rules in `.gitignore` (in addition to the root `.gitignore`, each package might also have its own `.gitignore`.)

The current script may work well, but developers could easily forget to update these YAML files when adding new files outside of `src` in the future. Using the grep rules from `.gitignore` will include some files that aren’t necessary for the RPM build, but since they’re not git-ignored, they’re probably small enough that it wouldn’t be a big deal.

We don’t have to do it now since the schedule is tight. Just keep it in mind and see if we can make this mechanism more robust in the future. 🙏

steven-chiu-bigstack (approved):
LGTM! 👍

.github/workflows/cube-cos-login-rpm.yaml
Eandalf-Bigstack: Got it, would look into this in the future, but yes, since we have a dedicated build output folder, the risk to wrap extraneous stuff into the rpm is pretty small.

PR #208, opened 2025-03-28
refactor(ci): make rpm source tarball creation more robust

Eandalf-Bigstack:
#### What type of PR is this?

- refactor

- Use `git checkout .` to move source code and make source tarball


steven-chiu-bigstack (approved):
Using `git checkout` looks like a brilliant idea! 🚀
Hope it works well. 🤞

PR #207, opened 2025-03-28
149 create tunings page

steven-chiu-bigstack:
Close #149 

# ⚠️ This PR is based on #184 ⚠️

#### What type of PR is this?

Feat

#### What this PR does / why we need it

Implemente the create tunings page.

#### Special notes for your reviewer

The current implementation only handles the creation of tunings. Edit existing tunings feature will be taken care of in #206.

#### Skeleton

chrome_ideruLY5qK

#### Create

https://github.com/user-attachments/assets/6aa89647-d783-44b3-a6d9-83d3aad8c06d

#### Go to Previous Step Through The Back Button

https://github.com/user-attachments/assets/bed708bf-792d-41b3-b478-5c2d91065702

PR #210, opened 2025-03-29
206 edit tunings page

### LINKED ISSUE #206: Page - Edit Tunings Page
[Mockup](https://www.figma.com/design/kRJ5MUAkMF0AYgXCeChtZn/COS-management-UI-Redesign-v.3?node-id=4880-4125&m=dev)

Image

Dependencies:

- #127 
- #149

steven-chiu-bigstack:
Close #206 

# ⚠️This PR is based on #207⚠️

#### What type of PR is this?

Feat

Implement the update tunings page.

The implementation is a bit complicated. In short:

1. Use Zustand store instead of query string to store the value of the edited tuning.
   - This is because there's no `id` in tunings, so we have to use tuning name (`key`) and hosts as the identifier. If we put those info in the URL, we might exceed the 2048-character limit as the host list grows.
   - Using [history.state](https://developer.mozilla.org/en-US/docs/Web/API/History/state) with React Router's [useLocation](https://api.reactrouter.com/v7/interfaces/react_router.Location.html) hook is an option, but since state is stored in the browser, navigating with the back and forward buttons will take users back to the edit tuning page with outdated information. This can cause users to see stale data (e.g., old value and old applied-hosts) if they revisit the page after making an update.
2. Edit non-modified tunings: Navigate to the edit tunings page with the selected spec name.
3. Edit modified tunings: Navigate to the edit tunings page with the selected spec name and applied host.

https://github.com/user-attachments/assets/3f3e7459-8ffa-4b76-9693-8c3463a2db84


raven-pan (approved):
Nice work!
I believe this is the best solution without the ID of the edit target.

PR #214, opened 2025-03-31
refactor(ci): use find to make mv more robust in rpm spec

Eandalf-Bigstack:
#### What type of PR is this?

- refactor

- Use find to make mv more robust in rpm spec


steven-chiu-bigstack (approved):
I'm not familiar with shell/bash scripts, but still LGTM! 💫 👍

PR #219, opened 2025-04-02
[ ✅ ready for review ] 147 implement Create Trigger page

### LINKED ISSUE #147: Page - Create Trigger Page
[Mockup](https://www.figma.com/design/kRJ5MUAkMF0AYgXCeChtZn/COS-management-UI-Redesign-v.3?node-id=3676-24675&m=dev)

Image

lizzy-liang-bigstack:
#### What type of PR is this?

- feature

##### 1) Select Template

- `When editing`: The table should display all available templates, but the others should be disabled.

https://github.com/user-attachments/assets/03be3de4-e5b6-46ac-a95d-d4e0c72fdd17

- `When creating`: The table should display all available templates, and all the rows are selectable.  It will be implemented in Phase 2.

Screenshot 2025-04-02 at 3 45 50 PM

##### 2) Select Events

- In Phase 1, attributes cannot be added, edited, or removed.

https://github.com/user-attachments/assets/e794602f-d3d4-4ecb-a1c0-6bd7f1c5f238

##### 3) Set Response

- ⚠️ How does the UI reflect changes after updating?

https://github.com/user-attachments/assets/91b82a1a-aa05-4d36-a0cb-848441a7e9e7

##### 4) Add Description

- ⚠️ The description field is not yet included in the API parameters.

https://github.com/user-attachments/assets/6ac68daa-a06e-441c-bb48-24e5d92ba5da

- close #147 

- It is based on PR #196 

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-ui-library/src/components/CosBasicTable/CosBasicTable.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosBasicTable/CosBasicTable.tsx
@@ -112,6 +112,7 @@ export const CosBasicTable = <Row extends CosTableRow>(
       <table className="w-full border-separate border-spacing-0">
         <thead>
           <tr>
+            {/*  */}
             {columns.map((column, index) => (
               <CosTableTh
                 key={`${column.property?.toString() ?? ''}-${index}`}
--- a/packages/cube-frontend-ui-library/src/components/CosBasicTable/CosBasicTable.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosBasicTable/CosBasicTable.tsx
@@ -4,6 +4,7 @@ import {
   PropsWithChildren,
   useMemo,
 } from 'react'
+import { cva } from 'class-variance-authority'
 import { twMerge } from 'tailwind-merge'
 import {
   computeRowClassName,
@@ -22,6 +23,20 @@ const tdBorderRadiusClass = twMerge(
   '[&:last-of-type>td:first-of-type]:rounded-bl-[5px]',
   '[&:last-of-type>td:last-of-type]:rounded-br-[5px]',
 )
+
+const tableRow = cva(['[&>td]:hover:bg-functional-hover-grey'], {
+  variants: {
+    isChecked: {
+      true: '[&>td]:bg-functional-hover-secondary [&>td]:hover:bg-[#ECF1FF]',
+    },
+    isDisabled: {
+      true: [
+        '[&>td]:bg-white [&>td]:hover:bg-white',
+        '[&>td]:text-functional-disable-text',
+      ],
+    },
+  },
+})
 
 export type CosBasicTableProps<Row extends CosTableRow> = PropsWithChildren<{
   rows: Row[]
@@ -89,9 +104,9 @@ export const CosBasicTable = <Row extends CosTableRow>(
       <tr
         key={row.id}
         className={twMerge(
-          '[&>td]:hover:bg-functional-hover-grey',
           tdBorderRadiusClass,
           computeRowClassName(rowClassName, row),
+          tableRow({ isChecked: !!row.checked, isDisabled: !!row.disabled }),
         )}
         onClick={() => onRowClick?.(row)}
       >
@@ -112,7 +127,6 @@ export const CosBasicTable = <Row extends CosTableRow>(
       <table className="w-full border-separate border-spacing-0">
         <thead>
           <tr>
-            {/*  */}
             {columns.map((column, index) => (
               <CosTableTh
                 key={`${column.property?.toString() ?? ''}-${index}`}
--- a/packages/cube-frontend-ui-library/src/components/CosBasicTable/cosTableUtils.ts
+++ b/packages/cube-frontend-ui-library/src/components/CosBasicTable/cosTableUtils.ts
@@ -4,6 +4,8 @@ import { CosTableColumnProps } from './rendering/CosTableColumn'
 
 export type CosTableRow = {
   id: string
+  checked?: boolean
+  disabled?: boolean
 }
 
 export type CosBatchActionTableRow = CosTableRow & {
--- a/packages/cube-frontend-ui-library/src/components/CosBatchActionTable/useCheckboxStatus.ts
+++ b/packages/cube-frontend-ui-library/src/components/CosBatchActionTable/useCheckboxStatus.ts
@@ -10,7 +10,8 @@ export const useCheckboxStatus = <Row extends CosTableRow>(
     if (selectedRowIds.length === 0) return false
 
     // If all rows' `id` are selected, return true
-    if (rows.every((row) => selectedRowIds.includes(row.id))) return true
+    const enabledRows = rows.filter((row) => !row.disabled)
+    if (enabledRows.every((row) => selectedRowIds.includes(row.id))) return true
 
     // If some rows are selected, return null (indeterminate state)
     return null
--- a/packages/cube-frontend-ui-library/src/components/CosBatchActionTable/useCheckboxStatus.ts
+++ b/packages/cube-frontend-ui-library/src/components/CosBatchActionTable/useCheckboxStatus.ts
@@ -1,0 +1,18 @@
+import { useMemo } from 'react'
+import { CosTableRow } from '../CosBasicTable/cosTableUtils'
+
+export const useCheckboxStatus = <Row extends CosTableRow>(
+  rows: Row[],
+  selectedRowIds: string[],
+): boolean | null => {
+  return useMemo(() => {
+    // If no rows are selected, return false
+    if (selectedRowIds.length === 0) return false
+
+    // If all rows' `id` are selected, return true
+    if (rows.every((row) => selectedRowIds.includes(row.id))) return true
+
+    // If some rows are selected, return null (indeterminate state)
+    return null
+  }, [rows, selectedRowIds])
+}
--- a/packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateActionButton.tsx
+++ b/packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateActionButton.tsx
@@ -9,32 +9,46 @@ export const TriggersCreateActionButton = () => {
   const {
     activeStep,
     setActiveStep,
+    isTemplateLoading,
+    selectedTemplateName,
     isTriggerLoading,
     enabled,
     attributes,
-    selectedTemplate,
     selectedEmails,
     selectedSlacks,
   } = useContext(TriggersCreateContext)
 
-  const { isUpdateLoading, handleTriggerUpdate, errorState } = useUpdateTrigger(
-    { enabled, attributes, selectedTemplate, selectedEmails, selectedSlacks },
-  )
-
-  const isLastStep = activeStep.serialNumber === 4
-
   const { isValid, errorMessage } = isFormValueValid(
     activeStep,
-    selectedTemplate,
+    selectedTemplateName,
     selectedEmails,
     selectedSlacks,
   )
 
+  const { isUpdateLoading, handleTriggerUpdate, errorState } = useUpdateTrigger(
+    {
+      isValid,
+      enabled,
+      attributes,
+      selectedTemplateName,
+      selectedEmails,
+      selectedSlacks,
+    },
+  )
+
+  const isLastStep = activeStep.serialNumber === 4
+
+  const isLoading = isTemplateLoading || isTriggerLoading
+
   const renderErrorMessage = () => {
-    const message =
-      errorMessage || errorState?.api?.msg || errorState?.native?.message
-    if (isTriggerLoading || !message) return null
-    return <div className="primary-body3 text-status-negative">{message}</div>
+    if (errorMessage || errorState?.api?.msg || errorState?.native?.message)
+      return (
+        <div className="primary-body3 text-status-negative">
+          {errorMessage || errorState?.api?.msg || errorState?.native?.message}
+        </div>
+      )
+
+    return null
   }
 
   const handleGoToNextStep = () => {
@@ -54,7 +68,7 @@ export const TriggersCreateActionButton = () => {
         size="md"
         type="primary"
         loading={isUpdateLoading}
-        disabled={!isValid || isTriggerLoading}
+        disabled={!!errorState}
         onClick={handleTriggerUpdate}
       >
         Update
@@ -70,7 +84,7 @@ export const TriggersCreateActionButton = () => {
         type="primary"
         usage="icon-right"
         Icon={ChevronRight}
-        loading={isTriggerLoading}
+        loading={isLoading}
         disabled={!isValid}
         onClick={handleGoToNextStep}
       >
--- a/packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateActionButton.tsx
+++ b/packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateActionButton.tsx
@@ -12,10 +12,10 @@ export const TriggersCreateActionButton = () => {
     isTemplateLoading,
     selectedTemplateName,
     isTriggerLoading,
-    enabled,
     attributes,
     selectedEmails,
     selectedSlacks,
+    description,
   } = useContext(TriggersCreateContext)
 
   const { isValid, errorMessage } = isFormValueValid(
@@ -28,11 +28,11 @@ export const TriggersCreateActionButton = () => {
   const { isUpdateLoading, handleTriggerUpdate, errorState } = useUpdateTrigger(
     {
       isValid,
-      enabled,
       attributes,
       selectedTemplateName,
       selectedEmails,
       selectedSlacks,
+      description,
     },
   )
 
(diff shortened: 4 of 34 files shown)


packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateGoBackButton.tsx
```diff
@@ -0,0 +1,34 @@
+import { useContext } from 'react'
+import { CosBackButton } from '@cube-frontend/ui-library'
+import { TriggersCreateContext } from './context'
+import { triggersCreateSteps } from './utils'
+import { useNavigate } from 'react-router'
+
+export const TriggersCreateGoBackButton = () => {
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateGoBackButton.tsx:7]: Besides using this button to go back to the previous step/page, it’d be great if the browser’s back button could do the same thing.

The Create/Update Tunings page handles this by storing the current step as a query param in the URL. That way, each step change pushes a new entry to `window.history`, so clicking the back button doesn’t immediately take the user back to the Triggers Listing page, meaning:

- The current step is a computed property based on a sepcific query param, e.g., `step=template` (no `step` param means we're at the first step)
- We can just call `history.back()` when `TriggersCreateGoBackButton` is clicked.

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateSteps.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { TriggersCreateContext } from './context'
+
+export const TriggersCreateSteps = () => {
+  const { activeStep } = useContext(TriggersCreateContext)
+
+  return (
+    <CosStepProcess isLoading={false}>
+      {triggersCreateSteps.map((step) => (
+        <CosStepProcess.Item
+          key={`create-${step.label}`}
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateSteps.tsx:13]: Nitpick: I think it's safe to use `step.label` as the key.

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateSteps.tsx
```diff
@@ -0,0 +1,21 @@
...
+          serialNumber={step.serialNumber}
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateSteps.tsx:14]: Nitpick: Since these `CosStepProcessItem`s are rendered within a for-loop, we can just use the index (`index + 1`) as `serialNumber`.

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateForm.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { TriggersStepTemplate } from './TriggersStepTemplate/TriggersStepTemplate'
+import { TriggersStepEvent } from './TriggersStepEvents/TriggersStepEvents'
+import { TriggersCreateActionButton } from './TriggersCreateActionButton'
+import { TriggersCreateContext } from './context'
+
+export const TriggersCreateForm = () => {
+  const { activeStep } = useContext(TriggersCreateContext)
+
+  const renderContent = () => {
+    if (activeStep.label === 'Select Template') {
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateForm.tsx:14]: Note: Code that depends heavily on UI labels is usually not i18n friendly.

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersStepResponse/ResponseEmailTable.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+type EmailTableType = GetTriggerResponseDataResponseEmailsInner & { id: string }
+
+const EmailTable = GetCosBatchActionTable<EmailTableType>()
+
+const mapToEmailTable = (
+  email: GetTriggerResponseDataResponseEmailsInner,
+): EmailTableType => ({
+  ...email,
+  id: email.email,
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersStepResponse/ResponseEmailTable.tsx:15]: There seems to be a schema mismatch in `GetTriggerResponseDataResponseEmailsInner`. The `email` field is actually named `address` in the API repsonse:

image

This has caused some bugs in the Set Response step:

chrome_rDYGcZ5ap1

We can create a PR in the openapi repo to fix this.

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useTemplateTable.ts
```diff
@@ -0,0 +1,82 @@
...
+   * while other template rows will be disabled.
+   *
+   * In Phase 2, a `Create a new trigger from template` feature will be implemented.
+   * At that time, there will be no query string in the URL, meaning no default `selectedTemplateName`.
+   * As a result, all rows in the table will be selectable, adn no rows will be pre-selected.
+   *
+   */
+  const [selectedTemplateName, setSelectedTemplateName] = useState<
+    string | null
+  >(urlTemplateName)
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useTemplateTable.ts:41]: I think we should check the validity of the `name` query param, even though no attributes will be selected when the trigger name does not exist. Otherwise if feels weird to be able to proceed the creation/update by knowing it'll definitely fail.

chrome_8SJIcQAUTx

The URL of the above example is `/events/triggers/create?name=qwer%20ty`.

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useTemplateTable.ts
```diff
@@ -0,0 +1,82 @@
...
+        .filter((template) => template.name !== selectedTemplateName)
+        .map((template) => template.id)
+    : []
+
+  const handleTemplateSelect = (selectedId: string) => {
+    const selectedTemplate = templateRows.find(
+      (template) => template.id === selectedId,
+    )
+    setSelectedTemplateName((prev) => {
+      return selectedTemplate ? selectedTemplate.name : prev
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useTemplateTable.ts:70]: Is this statement used to prevent users from deselecting a checkbox? If yes, I think we can make use of the 2nd argument of the `onCheckChange` on the batch action table to make the intent clearer:

```ts
const handleTemplateSelect = (rowId: string, checked: boolean) => {
  if (!checked) return

  const selectedTemplate = templateRows.find((row) => row.id === rowId)
  if (!selectedTemplate) return

  setSelectedTemplateName(selectedTemplate.name)
}
```

Alternatively, instead of putting `selectedTemplateName` in the context, we can just store the entire template (`selectedTemplate`) so we don't have to worry about the `!selectedTemplate` case (it shouldn't be happening, though.)

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+): GroupedAttribute | undefined => {
+  if (!arr) return
+  return arr.reduce((acc: GroupedAttribute, item) => {
+    if (!acc[item.name]) {
+      acc[item.name] = []
+    }
+    acc[item.name].push(item)
+    return acc
+  }, {})
+}
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/utils.ts:34]: (Just a note, no need to make any changes)
This can also be done with lodash; any of the following would work:

- `lodash.groupBy(arr, 'name')`
- `lodash.groupBy(arr, (item) => item.name)` -> This one is more type-safe.

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+ *   - `isValid`: `true` if the form values are valid for the given step, otherwise `false`.
+ *   - `errorMessage`: A string describing the validation error, or `undefined` if valid.
+ */
+export const isFormValueValid = (
+  activeStep: TriggersCreateStep,
+  selectedTemplateName: string | null,
+  selectedEmails: string[],
+  selectedSlacks: string[],
+): { isValid: boolean; errorMessage?: string } => {
+  switch (activeStep.label) {
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/utils.ts:47]: One downside of using `switch` and `if/else` is that if the possible values change during development, static checks such as `tsc` won’t catch any unhandled cases. Those issues only show up at runtime, which makes them easier to miss and more prone to bugs.

One way to work around this is to define a validation function map using `CreateStep` as the key. But we’ll need to update it to make it less tightly coupled to the UI. By the way, if you mean to use this page for create & update at the same time, perhaps `UpsertStep` would be a better name for it compared to `CreateStep`. For example:

```ts
enum UpsertStep {
  Template = 'template',
  Events = 'events',
  Response = 'response',
  Description = 'description',
}

const validationFnMap: Record<UpsertStep, (context: ContextValue) => boolean> => {
  template: (contextValue) => {
    return !!contextValue.selectedTemplateName
  },
  // ...other validation functions
}
```

This way when we add/remove/update a step from this page (i.e. when `UpsertStep` changes), `tsc` will notice `validationFnMap` doesn't match with specified type `Record<UpsertStep, (context: ContextValue) => boolean>`, helping us detect problems earlier (before shipping the application to production environment.) This also allows us to keep the validation code of each step separate (easier to read and maintain.)

This pattern can be applied to most of the scenarios, including content rendering for different steps.

steven-chiu-bigstack (commented):
Hi @lizzy-liang-bigstack 

Nice work! I left some comments for you to check. Let me know if you have any questions! 🙏

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts
```diff
@@ -0,0 +1,137 @@
...
+        } else if (trigger?.response?.slacks.length > 0) {
+          return [allSlacks[0]]
+        } else {
+          return []
+        }
+      })
+
+      setDescription(trigger.description)
+    }
+  }, [trigger, isTriggerLoading, allEmails, allSlacks])
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts:91]: Discussion: Do you think it make sense to still allow users to select disabled email recipients and slack channels while upserting a trigger? The code do select the first email/slack in the list when no row is enabled, so it's not that different.

Also, the current implementation selects all enabled emails and slack channels on initialization, while the mockup says to only check the first email by default:

{921A7D25-AD11-4DC7-B3AE-5C18F03C0E9C}

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts
```diff
@@ -0,0 +1,137 @@
...
+    () => trigger?.response?.slacks?.map((slack) => slack.url) ?? [],
+    [trigger],
+  )
+  const enabled = useMemo(() => trigger?.enabled ?? true, [trigger])
+
+  /**
+   * Sync form state with trigger data when it is available.
+   */
+  useEffect(() => {
+    if (!isTriggerLoading && trigger) {
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts:60]: Nitpick: Is it correct that we can remove the `!isTriggerLoading` check?

packages/cube-frontend-ui-library/src/components/CosBasicTable/CosBasicTable.tsx
```diff
@@ -89,9 +104,9 @@ export const CosBasicTable = <Row extends CosTableRow>(
       <tr
         key={row.id}
         className={twMerge(
-          '[&>td]:hover:bg-functional-hover-grey',
           tdBorderRadiusClass,
           computeRowClassName(rowClassName, row),
+          tableRow({ isChecked: !!row.checked, isDisabled: !!row.disabled }),
```
steven-chiu-bigstack, line 109: A few more comments after rebase:

It looks like this change should be discarded as it's already handled in [this commit](https://github.com/bigstack-oss/cube-cos-ui/pull/238/commits/65bbfe2c1cc9a062f4d42f1ef28486377b539eb6) from the batch action table [PR](https://github.com/bigstack-oss/cube-cos-ui/pull/238).

packages/cube-frontend-ui-library/src/components/CosBasicTable/cosTableUtils.ts
```diff
@@ -4,6 +4,8 @@ import { CosTableColumnProps } from './rendering/CosTableColumn'
 
 export type CosTableRow = {
   id: string
+  checked?: boolean
+  disabled?: boolean
```
steven-chiu-bigstack, line 8: These 2 lines should be removed.

packages/cube-frontend-ui-library/src/components/CosBatchActionTable/useCheckboxStatus.ts
```diff
@@ -0,0 +1,19 @@
+import { useMemo } from 'react'
+import { CosTableRow } from '../CosBasicTable/cosTableUtils'
+
+export const useCheckboxStatus = <Row extends CosTableRow>(
```
steven-chiu-bigstack, line 4: This file should be removed.

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts
```diff
@@ -0,0 +1,137 @@
...
+  const [selectedEmails, setSelectedEmails] = useState<string[]>([])
+  const [selectedSlacks, setSelectedSlacks] = useState<string[]>([])
+  const [description, setDescription] = useState<string>('')
+
+  /**
+   * Compute attributes – this field is temporarily non-editable in Phase 1.
+   */
+  const attributes = useMemo(
+    () =>
+      (trigger?.attributes ?? []) as GetTriggersResponseDataInnerAttributes[],
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts:43]: Do we use `as` because the API interface is incorrect? If that's true, we should fix it in the openapi yaml:

{95E38B77-EB98-4343-BC9C-8BA4BB16C0EA}

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts
```diff
@@ -0,0 +1,137 @@
...
+        } else if (trigger?.response?.emails?.length > 0) {
+          return [allEmails[0]]
+        } else {
+          return []
+        }
+      })
+
+      setSelectedSlacks(() => {
+        if (enabledSlacks.length > 0) {
+          return enabledEmails
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts:81]: This should be `enabledSlacks` instead.

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts
Comment (steven-chiu-bigstack) [reply packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts:91]: (Just a note, no need to make any changes)

This state initialization effect is a little bit long and not easy to read. A common way to mitigate this kind of issue is to combine these states and have a function that returns the initial state. For example:

```ts
type CreateTriggerFormValue = {
  selectedEmails: string[]
  selectedSlackUrls: string[]
  description: string
}

const [formValue, setFormValue] = useState<CreateTriggerFormValue>(() => ({
  selectedEmails: [],
  selectedSlackUrls: string[],
  description: '',
}))

useEffect(() => {
  if (trigger) {
    const initialFormValue = triggerToFormValue(trigger)
    setFormValue(initialFormValue)
  }
}, [trigger])

const triggerToFormValue = (trigger: GetTriggerResponseData): CreateTriggerFormValue => {
  // ...your implementation here.
}
```

It's okay to duplicate some code to make things easier.

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateActionButton.tsx
```diff
@@ -0,0 +1,105 @@
+import { useContext } from 'react'
+import ChevronRight from '@cube-frontend/ui-library/icons/monochrome/chevron_right.svg?react'
+import { CosButton } from '@cube-frontend/ui-library'
+import { useUpdateTrigger } from './useUpdateTrigger'
+import { isFormValueValid, triggersCreateSteps } from './utils'
+import { TriggersCreateContext } from './context'
+
+export const TriggersCreateActionButton = () => {
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateActionButton.tsx:8]: From the user's perspective, it makes sense to have a single button (component) that handles both stepping and creating on the Create Triggers page. But that also means the button has to be “smart,” which adds complexity to the code.

I've considered doing the same thing on the Create/Update Tunings page, but eventually went with a dumber approach—placing separate buttons within each step. That way, the `onClick` and form validation logic stay encapsulated in each step, making things feel more modular and easier to manage.

The current code works, but if we ever add more steps or constraints to the creation flow, it probably won’t scale well. Since the trigger creation flow is still simple, I think it’s okay to keep the current implementation for now. 👌

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateGoBackButton.tsx
Comment (lizzy-liang-bigstack) [reply packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/TriggersCreateGoBackButton.tsx:7]: Thanks for the suggestion! I totally agree — using the browser’s back button for step navigation makes the UX feel a lot more natural. I’ve gone ahead and implemented the same approach as the Create/Update Tunings page by tracking the step in the query params (e.g., `step=template`) and updating the URL accordingly on each step change.

packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts
Comment (lizzy-liang-bigstack) [reply packages/cube-frontend-web-app/src/pages/events/triggers/[create]/_components/useCreateTriggerForm.ts:91]: Form my understanding, the `enabled` value in the Emails and Slacks data determines whether each checkbox is checked.

The purpose of the table is to allow users to modify the `enabled` status, so they can re-enable a currently disabled email or Slack option - if this makes sense.

PR #238, opened 2025-04-11
46 ui library - CosBatchActionTable

### LINKED ISSUE #46: UI Library - Table Batch Action
[Mockup](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=82-2772&m=dev)

Dependencies:

- #43

lizzy-liang-bigstack:
#### What type of PR is this?

- Feature

- Create `CosBatchActionTable` component

##### 1) Default

https://github.com/user-attachments/assets/c22adeae-0637-431d-af8d-a3ef82e2e605

##### 2) Without checkout on the header

- It is not in the design system but a variant implementation on `Create Trigger` page

https://github.com/user-attachments/assets/6d83a1d6-faa2-4f61-8a61-765bc2654ff8

##### 3) Skeleton

Screenshot 2025-04-02 at 12 46 11 PM

- Close #46 

> Note

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-ui-library/src/components/CosBasicTable/cosTableUtils.ts
+++ b/packages/cube-frontend-ui-library/src/components/CosBasicTable/cosTableUtils.ts
@@ -4,6 +4,9 @@ import { CosTableColumnProps } from './rendering/CosTableColumn'
 
 export type CosTableRow = {
   id: string
+}
+
+export type CosBatchActionTableRow = CosTableRow & {
   checked?: boolean
   disabled?: boolean
 }
--- a/packages/cube-frontend-ui-library/src/components/CosBasicTable/cosTableUtils.ts
+++ b/packages/cube-frontend-ui-library/src/components/CosBasicTable/cosTableUtils.ts
@@ -4,6 +4,8 @@ import { CosTableColumnProps } from './rendering/CosTableColumn'
 
 export type CosTableRow = {
   id: string
+  checked?: boolean
+  disabled?: boolean
 }
 
 export const COS_TABLE_COLUMN_SYMBOL = Symbol('CosTableColumn')
--- a/packages/cube-frontend-ui-library/src/components/CosBatchActionTable/CosBatchActionTable.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosBatchActionTable/CosBatchActionTable.tsx
@@ -1,0 +1,104 @@
+import { useCallback, useMemo } from 'react'
+import {
+  CosBasicTable,
+  CosBasicTableProps,
+} from '../CosBasicTable/CosBasicTable'
+import { CosTableRow } from '../CosBasicTable/cosTableUtils'
+import { CreateCosTableColumn } from '../CosBasicTable/rendering/CosTableColumn'
+import { CosCheckbox } from '../CosCheckbox/CosCheckbox'
+import { CosSkeleton } from '../CosSkeleton/CosSkeleton'
+import { useCheckboxStatus } from './useCheckboxStatus'
+
+export type CosBatchActionTableProps<Row extends CosTableRow> =
+  CosBasicTableProps<Row> & {
+    selectedRowIds: string[]
+    disabledRowIds?: string[]
+    onCheckChange: (rowId: string) => void
+  } & (
+      | {
+          showHeaderCheckbox: true
+          onAllCheckChange: () => void
+        }
+      | { showHeaderCheckbox: false }
+    )
+
+const CosBatchActionTable = <Row extends CosTableRow>(
+  props: CosBatchActionTableProps<Row>,
+) => {
+  const {
+    children,
+    rows,
+    selectedRowIds,
+    disabledRowIds = [],
+    onCheckChange,
+    showHeaderCheckbox,
+    ...restProps
+  } = props
+
+  const isRowSelected = useCallback(
+    (rowId: string) => selectedRowIds.includes(rowId),
+    [selectedRowIds],
+  )
+  const isRowDisabled = useCallback(
+    (rowId: string) => disabledRowIds.includes(rowId),
+    [disabledRowIds],
+  )
+
+  const convertedRows: Row[] = useMemo(
+    () =>
+      rows.map((row) => ({
+        ...row,
+        checked: isRowSelected(row.id),
+        disabled: isRowDisabled(row.id),
+      })),
+    [isRowDisabled, isRowSelected, rows],
+  )
+
+  const checkboxStatus = useCheckboxStatus(convertedRows, selectedRowIds)
+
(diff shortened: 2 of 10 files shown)


packages/cube-frontend-ui-library/src/components/CosBasicTable/cosTableUtils.ts
```diff
@@ -4,6 +4,8 @@ import { CosTableColumnProps } from './rendering/CosTableColumn'
 
 export type CosTableRow = {
   id: string
+  checked?: boolean
+  disabled?: boolean
 }
```
steven-chiu-bigstack, line 12: I think it makes more sense to separate `checked` and `disabled` from `CosTableRow` (even though they're both optional.) The main reason is that this type is also used by `CosBasicTable`, but they only effect batch actions tables. Users should not be aware of `checked` and `disabled` when using `CosBasicTable`:

```ts
export type CosTableRow = {
  id: string
}

export type CosBatchActionTableRow = CosTableRow & {
  checked?: boolean
  disabled?: boolean
}
```

packages/cube-frontend-ui-library/src/components/CosBatchActionTable/CosBatchActionTable.tsx
```diff
@@ -0,0 +1,104 @@
...
+        checked={checkboxStatus}
+        onChange={() => onAllCheckChange()}
+      />
+    )
+  }
+
+  return (
+    <CosBasicTable {...restProps} rows={convertedRows}>
+      <CosBasicTable.Column label={renderHeaderCheckbox()} property="id">
+        {(id) => {
```
steven-chiu-bigstack: Row item is provided as the second argument of the cell rendering function, so we should be able to use `row.disabled` and `row.checked` directly without calling `isRowDisabled` and `isRowSelected` again here:

```tsx
{(id, row) => { ... }}
```

packages/cube-frontend-ui-library/src/components/CosBatchActionTable/useCheckboxStatus.ts
```diff
@@ -0,0 +1,19 @@
...
+  rows: Row[],
+  selectedRowIds: string[],
+): boolean | null => {
+  return useMemo(() => {
+    // If no rows are selected, return false
+    if (selectedRowIds.length === 0) return false
+
+    // If all rows' `id` are selected, return true
+    const enabledRows = rows.filter((row) => !row.disabled)
+    if (enabledRows.every((row) => selectedRowIds.includes(row.id))) return true
```
steven-chiu-bigstack: Nitpick: We can convert `selectedRowIds` to a `Set` to reduce time complexity from `M * N` to `M + N`.

packages/cube-frontend-ui-library/src/components/CosBatchActionTable/useCheckboxStatus.ts
```diff
@@ -0,0 +1,19 @@
+import { useMemo } from 'react'
+import { CosTableRow } from '../CosBasicTable/cosTableUtils'
+
+export const useCheckboxStatus = <Row extends CosTableRow>(
```
steven-chiu-bigstack: Nitpick: We can rename this hook to `useHeaderCheckboxStatus` to align with its usage.

packages/cube-frontend-ui-library/src/components/CosBatchActionTable/CosBatchActionTable.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+export type CosBatchActionTableProps<Row extends CosTableRow> =
+  CosBasicTableProps<Row> & {
+    selectedRowIds: string[]
+    disabledRowIds?: string[]
+    onCheckChange: (rowId: string) => void
+  } & (
+      | {
+          showHeaderCheckbox: true
+          onAllCheckChange: () => void
```
steven-chiu-bigstack: Since the prop name has a `Change` suffix, it's recommended to provide the new value (or the change event) to the event handler.

packages/cube-frontend-ui-library/src/stories/components/CosTables/BatchAction/CosBatchActionTable.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+    [],
+  )
+
+  useEffect(() => {
+    if (isAllSelected) {
+      setSelectedRowIds(allTemplateIds)
+    } else {
+      setSelectedRowIds([])
+    }
+  }, [allTemplateIds, isAllSelected])
```
steven-chiu-bigstack: We can remove the `isAllSelected` state and this `useEffect` if the changed value/event is emitted from `onAllCheckChange`.

steven-chiu-bigstack (commented):
Hi @lizzy-liang-bigstack 

Looks good! 👍 
The main feature works great. Just had a couple of thoughts around type safety and ease of use.

packages/cube-frontend-ui-library/src/components/CosBasicTable/cosTableUtils.ts
lizzy-liang-bigstack: @steven-chiu-bigstack Thanks for the feedbacks! 🙇‍♀️🙇‍♀️
  
Regarding your suggestion: I originally included the `checked` and `disabled` properties in the `CosTableRow` type because they were used to determine whether a row was selected or disabled, which directly affected its styling.
  
Now that the `CosTableRow` type no longer includes these two properties, I’ve updated the implementation to instead pass `selectedRowSet` and `disabledRowSet` into the `CosBasicTable` component. This way, when rendering data rows, the component can still determine whether a row is selected or disabled based on these sets.
  
I hope this adjustment aligns better with the suggested direction—please let me know if you have any further thoughts!

steven-chiu-bigstack: Hi @lizzy-liang-bigstack 

We’re very close, awesome work! 👏👏👏

One last thing: by adding `selectedRowSet` and `disabledRowSet` to `CosBasicTable`, users might try to pass them in, thus accidentally (or intentionally) apply the selected-hover or disabled style to a `CosBasicTable`.

Since those sets are just used to determine the (hover) class name, we can make use of the `rowClassName` prop in `CosBasicTable` to keep the sets internal to `CosBatchActionTable`.

I've pushed a [commit](https://github.com/bigstack-oss/cube-cos-ui/pull/238/commits/c6f43d5ccf3b5a90ef847b9cbdbed66f6bc0022a) to refactor it. Let me know if you have any questions! 🙏

steven-chiu-bigstack (approved):

PR #241, opened 2025-04-14
190 chart page missing features

steven-chiu-bigstack:
#### What type of PR is this?

Feature

#### What this PR does / why we need it

Integrate Grafana link API to Home Chart page.

#### Which issue(s) this PR fixes

#190 

#### Special notes for your reviewer

- This PR only handles the Grafana links in #190 
- The "Extra Information" panel at the bottom is only visible when the related feature is enabled from COS/Grafana.
   - image

##### Links

https://github.com/user-attachments/assets/9b15a7b6-b701-422d-a957-9b5c06426f3d

PR #245, opened 2025-04-15
199 node details page

### LINKED ISSUE #199: Page - Node Details Page
[Mockup](https://www.figma.com/design/kRJ5MUAkMF0AYgXCeChtZn/COS-management-UI-Redesign-v.3?node-id=3281-14729&m=dev)

steven-chiu-bigstack:
# ⚠️ This PR is based on #241 ⚠️

#### What type of PR is this?

Feat

Implement the Node Details page.

Fixes #199

- It is confirmed that the **SSH Console** and **all ipmi related features** are excluded from phase 1.
- The amount of devices is higher than expected in practical scenarios, so I added a pagination below the device table.
- As stated in the mockup comment, we can ignore the "Status" column in the device table.
   - {CC005FE9-B5B0-4A99-AA36-9C173E2C09A5}
- The ellipsis menu in the device table is also excluded from phase 1.

https://github.com/user-attachments/assets/ae6b1a64-632e-4d7d-8860-63ade8538b3a

### DIFF
--- a/packages/cube-frontend-web-app/src/components/CopyButton.tsx
+++ b/packages/cube-frontend-web-app/src/components/CopyButton.tsx
@@ -10,7 +10,12 @@ export type CopyButtonProps = {
 }
 
 export const CopyButton = (props: CopyButtonProps) => {
-  const { copyContent: copyText, ...restProps } = props
+  const { copyContent: copyText, onClick: onClickProp, ...restProps } = props
+
+  const onClick = (e: MouseEvent<HTMLElement>): void => {
+    navigator.clipboard.writeText(copyText)
+    onClickProp?.(e)
+  }
 
   return (
     <CosButton
@@ -19,7 +24,7 @@ export const CopyButton = (props: CopyButtonProps) => {
       size="sm"
       Icon={CopyIcon}
       {...restProps}
-      onClick={() => navigator.clipboard.writeText(copyText)}
+      onClick={onClick}
     />
   )
 }
--- a/packages/cube-frontend-web-app/src/components/CopyButton.tsx
+++ b/packages/cube-frontend-web-app/src/components/CopyButton.tsx
@@ -1,12 +1,16 @@
 import { CosButton } from '@cube-frontend/ui-library'
 import CopyIcon from '@cube-frontend/ui-library/icons/monochrome/copy.svg?react'
+import { MouseEvent } from 'react'
 
 export type CopyButtonProps = {
   copyContent: string
+  onMouseEnter?: (e: MouseEvent<HTMLElement>) => void
+  onMouseLeave?: (e: MouseEvent<HTMLElement>) => void
+  onClick?: (e: MouseEvent<HTMLElement>) => void
 }
 
 export const CopyButton = (props: CopyButtonProps) => {
-  const { copyContent: copyText } = props
+  const { copyContent: copyText, ...restProps } = props
 
   return (
     <CosButton
@@ -14,6 +18,7 @@ export const CopyButton = (props: CopyButtonProps) => {
       type="ghost"
       size="sm"
       Icon={CopyIcon}
+      {...restProps}
       onClick={() => navigator.clipboard.writeText(copyText)}
     />
   )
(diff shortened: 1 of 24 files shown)


packages/cube-frontend-web-app/src/components/CopyButton.tsx
```diff
@@ -1,20 +1,30 @@
 import { CosButton } from '@cube-frontend/ui-library'
 import CopyIcon from '@cube-frontend/ui-library/icons/monochrome/copy.svg?react'
+import { MouseEvent } from 'react'
 
 export type CopyButtonProps = {
   copyContent: string
+  onMouseEnter?: (e: MouseEvent<HTMLElement>) => void
+  onMouseLeave?: (e: MouseEvent<HTMLElement>) => void
```
lizzy-liang-bigstack, line 8: (Just a question) 
  
I’m curious — why do we need onMouseEnter and onMouseLeave in the props? 🤔 It seems like we’re not using them anywhere.

packages/cube-frontend-web-app/src/components/NodeTable/NodeTable.tsx
```diff
@@ -26,7 +29,15 @@ export const NodeTable = (props: NodeTableProps) => {
         label="Hostname"
         property="hostname"
         emphasize={true}
-      />
+      >
+        {(hostname) => (
+          <Link className="block w-fit" to={`/nodes/${hostname}`}>
+            <CosHyperlink variant="text-only" onClick={noop}>
```
lizzy-liang-bigstack, line 35: (Just a question)
   
Since the the `onClick` prop is optional and the `noop` function returns `undefined`, I wonder why we are passing it at all? Wouldn't it be cleaner to simply omit the prop?

packages/cube-frontend-web-app/src/pages/node/[name]/_components/nodeChartsUtils.ts
```diff
@@ -0,0 +1,198 @@
...
+            return {
+              borderColor: cpuLineColor,
+              backgroundColor: cpuLineColor,
+            }
+          },
+          label: (tooltipItem) => {
+            const value = tooltipItem.parsed.y.toFixed(2)
+            const unitText = unit === 'percentage' ? '%' : unit
+            return `${value}${unitText}`
+          },
```
Comment (lizzy-liang-bigstack) [inline packages/cube-frontend-web-app/src/pages/node/[name]/_components/nodeChartsUtils.ts:111]: (Note)
  
I think the way we are handling the tooltip style here is clean and straightforward. It could be a enhancement to consider unifying the tooltip style across other charts :)

lizzy-liang-bigstack (approved):

packages/cube-frontend-web-app/src/components/CopyButton.tsx
steven-chiu-bigstack: Good question! This is because we wrap `CopyButton` with `CosTooltip`, which relies on `onMouseEnter`, `onMouseLeave`, and `onClick` to work properly (passed via `cloneElement`):

```tsx
<CosTooltip hoverContent={...} clickContent={...}>
  <CopyButton>
</CosTooltip>
```

Unlike Vue, React doesn’t automatically pass props to the child component’s root node, so we have to explicitly define and bind them ourselves.

We should find a way to make this more obvious, perhaps showing an error when the children does not have those props defined.

packages/cube-frontend-web-app/src/components/NodeTable/NodeTable.tsx
steven-chiu-bigstack: This is because `CosHyperlink` expects either `href` or `onClick` to be provided, otherwise, it logs a warning to the console:

image

The warning is reasonable since it’s designed to be framework-agnostic, but we can consider removing it if it becomes troublesome in most cases.

packages/cube-frontend-web-app/src/pages/node/[name]/_components/nodeChartsUtils.ts
Comment (steven-chiu-bigstack) [reply packages/cube-frontend-web-app/src/pages/node/[name]/_components/nodeChartsUtils.ts:111]: I copied them from Raven's Home Chart page implementation. 🐈   
Actually, this make me notice an issue here. I was meant to write ``const unitText = unit === 'percentage' ? '%' : ` ${unit}`;`` so that it can correctly display `35%` or `35 MiB` (align with Raven's Home Chart.)
I'll fix this later, thanks!

PR #246, opened 2025-04-15
refactor: update TimeRange component to support dynamic time ranges

lizzy-liang-bigstack:
#### What type of PR is this?

- Refactor

- Update the original `TimeRangeDropdown` to support custom time range options

Fixes #

> Note

#### Additional documentation

```docs

```


steven-chiu-bigstack (commented):
Hi @lizzy-liang-bigstack 

Thank you so much for taking care of this one! 🙏 
For the implementation, I think we could reuse `GetHealthHistoryPastEnum` from the API SDK to:

1. Reduce duplicate code by defining labelMap and pastMap once in the time range utils.
   - The only difference between them is the count. The labels and past functions are the same.
2. Improve type safety
   - Currently, `useTimeRange` and `<TimeRangeDropdown>` accept any string. But type mismatch will eventually be caught by TS when attempting to use it to make API calls. That said, it’d still be ideal to catch the error as early as possible.

{4D25EC3D-722A-46D2-BEDC-95E66D956147}

Either `GetHealthHistoryPastEnum` or `GetServiceHealthHistoryPastEnum` would work here since they both cover all the possible options. For example:

```ts
const { timeRange, past, onTimeRangeChange } = useTimeRange({
  // Must be an array of `GetHealthHistoryPastEnum`.
  includes: ['1h', '24h', '7d', '14d'],
  // Must be a value presented in the `includes` array.
  defaultValue: '1h',
})
```

Do you think this make sense? 🤔

packages/cube-frontend-web-app/src/components/TimeRangeDropdown/useTimeRange.ts
```diff
@@ -1,32 +1,31 @@
...
+  timeRangePastMap: Record<T, K>
 }
 
-const timeRangePastMap: Record<TimeRange, GetHealthHistoryPastEnum> = {
-  last30Days: '30d',
-  last14Days: '14d',
-  last7Days: '7d',
-  last24Hours: '24h',
-  lastHour: '1h',
+export type UseTimeRange<T extends string, K extends string> = {
```
steven-chiu-bigstack: If we were to stick to the current implementation, it is recommended to rename `T` and `K` to make them easier to understand. `TimeRange` and `Past` sounds fine:

```ts
type UseTimeRange<TimeRange extends string, Past extends string> = {...}
```

lizzy-liang-bigstack: Hi @steven-chiu-bigstack, thank you for the great suggestion!

I noticed that the current implementation requires two separate mappings to derive the values we need:
1. Mapping `timeRange` to labels
2. Mapping `timeRange` to `past` values

In this refactored version, I suggest using the value directly from `GetHealthHistoryPastEnum`. This would eliminate the need to map `timeRange` to a separate `past` value. Instead, the selected `timeRange` can be used directly in the API request.
 
The rest of the logic remains consistent with the existing implementation, but this update makes sure that the `includes` values passed into the `useTimeRange` hook come from `GetHealthHistoryPastEnum`.

packages/cube-frontend-web-app/src/components/TimeRangeDropdown/timeRangeUtils.ts
```diff
@@ -0,0 +1,19 @@
+import { GetHealthHistoryPastEnum } from '@cube-frontend/api'
+
+export const timeRanges = [
+  '30d',
+  '14d',
+  '7d',
+  '24h',
+  '1h',
+] as const satisfies readonly GetHealthHistoryPastEnum[]
```
steven-chiu-bigstack: Nitpick: We could simplify the `timeRanges` declaration to the following:

```ts
const timeRanges = Object.values(GetHealthHistoryPastEnum)
```

There's a difference between them, but either one works fine for our use case.

packages/cube-frontend-web-app/src/components/TimeRangeDropdown/useTimeRange.ts
```diff
@@ -1,32 +1,31 @@
...
+import { useCallback, useContext, useEffect, useState } from 'react'
+import { DataCenterContext } from '@cube-frontend/web-app/context/DataCenterContext'
+import { TimeRange } from './timeRangeUtils'
 
-export type UseTimeRange = {
-  now: Dayjs
-  timeRange: TimeRange
-  past: GetHealthHistoryPastEnum
-  onTimeRangeChange: (newTimeRange: TimeRange) => void
+export type UseTimeRangeOption<T extends readonly TimeRange[]> = {
```
steven-chiu-bigstack, line 6: I don't think `readonly` here does much to help check the array immutability, but it can still serve as a hint that the argument isn't meant to be modified. 👌

packages/cube-frontend-web-app/src/components/TimeRangeDropdown/useTimeRange.ts
```diff
@@ -1,32 +1,31 @@
...
-  last30Days: '30d',
-  last14Days: '14d',
-  last7Days: '7d',
-  last24Hours: '24h',
-  lastHour: '1h',
+export type UseTimeRange<T extends readonly TimeRange[]> = {
+  now: Dayjs
+  timeRange: T[number]
+  onTimeRangeChange: (newTimeRange: T[number]) => void
 }
```
steven-chiu-bigstack, line 20: To make the time range dropdown a bit easier to work with, we can expose the `includes` array from the `useTimeRange` hook (in the return value). That way, we don’t need to manually declare `const timeRanges = [...] as const` and pass it around to both the hook and the dropdown. For example:

{EE2DB80B-D59D-4B2B-A2A7-B855F50DFAE9}

packages/cube-frontend-web-app/src/pages/home/health/healthTimeRangeUtils.ts
```diff
@@ -0,0 +1,3 @@
+export const healthTimeRanges = ['30d', '14d', '7d', '24h', '1h'] as const
```
steven-chiu-bigstack: Nitpick: We can add `satisfies TimeRange[]` here to ensure type safety from the root.

packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeEvents.tsx
```diff
@@ -30,7 +31,10 @@ export const NodeEvents = (props: NodeEventsProps) => {
 
   const { name: dataCenter } = useContext(DataCenterContext)
 
-  const { timeRange, past, onTimeRangeChange } = useTimeRange()
+  const { timeRange, onTimeRangeChange } = useTimeRange({
+    includes: chartTimeRanges,
+    defaultValue: '1h',
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeEvents.tsx]: The default time range for events table should be `24h`.

steven-chiu-bigstack (approved):
Excellent work! I've pushed a commit to deal with some refactoring mentioned in my comments. 🙏

PR #254, opened 2025-04-18
feat(ui-library): view details table

### LINKED ISSUE #248: UI Library - Table View Details
[Mockup](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=4024-8196&m=dev)

steven-chiu-bigstack:
#### What type of PR is this?

Feat

Implement the `CosViewDetailsTable` component.

#248

https://github.com/user-attachments/assets/690ab2b4-bcbc-4467-a07e-2926d87876bf

1. Checkbox functionality is not implemented yet since there’s no actual use case at the moment. We can deal with that later.
2. The current implementation is a little bit different from the mockup:
   - Mockup:
      - {088D3A87-E0BD-4C08-B003-13FCE547F3B8}
   - Current implementation:
      - image
   - As shown above, the expand button is designed to be embedded in the first column, with no extra column reserved for it. But in the current implementation, it is placed in a separate column at the start of the table. While it's possible to match the design, it would require more complex logic (modifying nested children), so I’ve decided to skip it for now.

### DIFF
--- a/packages/cube-frontend-ui-library/src/components/CosBasicTable/CosBasicTable.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosBasicTable/CosBasicTable.tsx
@@ -1,6 +1,7 @@
 import {
   ComponentProps,
   ComponentType,
+  Fragment,
   PropsWithChildren,
   useMemo,
 } from 'react'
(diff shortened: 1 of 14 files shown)


raven-pan (approved):
LGTM.
Thanks for your help!

PR #265, opened 2025-04-22
Feat/182 invalid license nagging

raven-pan:
## Main Feature

Implemented SideBar and Top License Nagging

Screenshot 2025-04-22 at 1 54 10 PM

### Notes

1. The Top License Nagging only appears when the user visists License Page and can be closed by clicking the close button.
2. The SideBar License Nagging displays title according to this [priority](https://docs.google.com/spreadsheets/d/1JDzkw_Ur3tZc0uZdRU6z8AQ1GfWAh12jAgaB33zb_gc/edit?pli=1&gid=).
3. The Top License Nagging displays description order according to this [priority](https://docs.google.com/spreadsheets/d/1JDzkw_Ur3tZc0uZdRU6z8AQ1GfWAh12jAgaB33zb_gc/edit?pli=1&gid=).

### Nagging

1.) Implemented the close feature for Top Nagging

Screenshot 2025-04-22 at 2 12 30 PM

2.) Implement description list for both Top and SideBar Nagging

Screenshot 2025-04-22 at 2 13 09 PM

Screenshot 2025-04-22 at 2 13 37 PM

### SideBar

1.) Displays Nagging when the parent  component passes the `naggingProps` to `CosSideBar`.

image

## Testing Plan

1.) ui-library: Check `Nagging` and `SideBar` components in Storybook.
2.) web-app: Mock `nodeLicenseStatus` in `DataCenterProvider.tsx`

image

### DIFF
--- a/packages/cube-frontend-ui-library/src/components/CosNagging/CosNagging.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosNagging/CosNagging.tsx
@@ -1,26 +1,34 @@
+import { isArray } from 'lodash'
 import WarningFilled from '../../components/CosIcon/monochrome/warning_filled.svg?react'
 import WarningAltFilled from '../../components/CosIcon/monochrome/warning_alt_filled.svg?react'
+import XSmallIcon from '@cube-frontend/ui-library/icons/monochrome/x_small.svg?react'
 import { PropsWithClassName } from '@cube-frontend/utils'
 import { cva } from 'class-variance-authority'
 import { twMerge } from 'tailwind-merge'
-import { CosHyperlink } from '../CosHyperlink/CosHyperlink'
+import { CosHyperlink, CosHyperlinkProps } from '../CosHyperlink/CosHyperlink'
 import { SvgElement } from '../CosIcon/CosIcon'
 
 export type CosNaggingType = 'error' | 'warning'
 
 export type CosNaggingVariant = 'sidebar' | 'top'
 
+export type CosNaggingLink = Pick<CosHyperlinkProps, 'href' | 'onClick'> & {
+  text: CosHyperlinkProps['children']
+}
+
 export type CosNaggingProps = PropsWithClassName & {
   type: CosNaggingType
   title: string
+  description?: string | string[]
+  link?: CosNaggingLink
 } & (
-    | { variant: Extract<CosNaggingVariant, 'sidebar'>; description?: string }
-    | { variant: Extract<CosNaggingVariant, 'top'> }
-  ) & { link?: { href: string; text: string } }
+    | { variant: Extract<CosNaggingVariant, 'sidebar'> }
+    | { variant: Extract<CosNaggingVariant, 'top'>; onClose?: () => void }
+  )
 
 const nagging = cva(
   [
-    'flex gap-[6px] self-start rounded-md border bg-yellow-50 p-3',
+    'flex flex-col gap-[6px] self-start rounded-md border bg-yellow-50 p-3',
     'shadow-[0_0_2px_0_rgba(0,0,0,0.2)]',
   ],
   {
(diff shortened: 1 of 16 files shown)


packages/cube-frontend-ui-library/src/components/CosNagging/CosNagging.tsx
```diff
@@ -31,7 +39,7 @@ const nagging = cva(
       },
       variant: {
         sidebar: 'flex-col',
-        top: '',
+        top: 'flex-col',
       },
```
lizzy-liang-bigstack: Since the `sidebar` and `top` variants share the same style, perhaps it would be simpler to move `flex-col` to the common styles?

packages/cube-frontend-ui-library/src/components/CosNagging/CosNagging.tsx
```diff
@@ -108,9 +148,12 @@ export const CosNagging = (props: CosNaggingProps) => {
...
-        {renderTitle()}
+      <div className="flex items-center justify-between">
+        <div className="flex items-start gap-2">
+          {icon}
+          {renderTitle()}
+        </div>
+        {renderCloseIcon()}
       </div>
       {renderDescription()}
       {renderBottomLink()}
```
lizzy-liang-bigstack, line 166: It seems there’s extra space left at the bottom in the following cases:

- `Top` variant: when there’s no `description` in the content.
Screenshot 2025-04-22 at 2 59 48 PM

Screenshot 2025-04-22 at 3 24 56 PM

- `Sidebar` variant: when the content only contains a `title`.

Screenshot 2025-04-22 at 3 26 47 PM

packages/cube-frontend-web-app/src/pages/maintenance/MaintenanceLayout.tsx
```diff
@@ -1,12 +1,13 @@
...
+import { CosTabs } from '@cube-frontend/ui-library'
 import { links } from './links'
+import { TopLicenseNagging } from './_components/TopLicenseNagging'
 
 export const MaintenanceLayout = () => {
   const location = useLocation()
-
   return (
     <div className="flex flex-col gap-y-4">
+      <TopLicenseNagging />
```
lizzy-liang-bigstack, line 10: Just a quick question — since the nagging is placed at the same level as the tabs, it doesn’t re-appear once it’s closed when switching between tabs, only when navigating between pages. Is this the expected behaviour?

https://github.com/user-attachments/assets/2cf8e03e-93fd-49ac-bba9-fb04021be608

lizzy-liang-bigstack (commented):
I noticed that the appearance of the `Nagging` component is slightly different from the mockup — specifically, the position of the call-to-action button appears to be placed differently. 

It seems we don’t currently have a “multiple descriptions” layout defined in the design guidelines. The current implementation actually aligns more closely with the original design guideline, which I think is preferable. Maybe we could double-check with the designer to see if the mockup should be updated to reflect this. That said, I think it's totally fine to leave it as-is for now. Thank you!

Screenshot 2025-04-22 at 4 25 47 PM

Screenshot 2025-04-22 at 4 26 58 PM

Screenshot 2025-04-22 at 3 45 23 PM

packages/cube-frontend-ui-library/src/components/CosNagging/CosNagging.tsx
raven-pan: Thanks!
I will move it to common styles.

lizzy-liang-bigstack (approved):

packages/cube-frontend-web-app/src/pages/maintenance/MaintenanceLayout.tsx
raven-pan: Nice catch!
The Top Nagging should reappear once the user switches back to the License Tab.
I will fix it.

raven-pan: There are some differences between the License Page mockup and the Nagging design guidelines.

For now, I'll stick to the design guidelines.
I'll discuss it with the designers if we have time.
Thanks!

raven-pan:
> I noticed that the appearance of the `Nagging` component is slightly different from the mockup — specifically, the position of the call-to-action button appears to be placed differently.
> 
> It seems we don’t currently have a “multiple descriptions” layout defined in the design guidelines. The current implementation actually aligns more closely with the original design guideline, which I think is preferable. Maybe we could double-check with the designer to see if the mockup should be updated to reflect this. That said, I think it's totally fine to leave it as-is for now. Thank you!
> 
> Screenshot 2025-04-22 at 4 25 47 PM Screenshot 2025-04-22 at 4 26 58 PM Screenshot 2025-04-22 at 3 45 23 PM

There are some differences between the License Page mockup and the Nagging design guidelines.

For now, I'll stick to the design guidelines.
I'll discuss it with the designers if we have time.
Thanks!

packages/cube-frontend-ui-library/src/components/CosNagging/CosNagging.tsx
raven-pan: I'll fix it, thanks!

packages/cube-frontend-web-app/src/pages/maintenance/MaintenanceLayout.tsx
raven-pan: I fixed it by using Zustand.

Screenshot 2025-04-24 at 5 45 54 PM

PR #270, opened 2025-04-23
229 fix flickering issue on the “Create Support Files” button when the modal is closing

### LINKED ISSUE #229: [Bug]:  Node Page Create Support Files Modal Button Flickers When the User click Create Support Files Button
## Describe the bug

https://github.com/user-attachments/assets/0b31f3d5-f81a-4104-a39c-dd1b547f7930

## To Reproduce

1. Go to the `Node List Page`.
2. Click the `Create support files` button, it will open a modal.
3. Click the `Create support files` button in the modal.
4. While the modal is in its closing transition, the `Create support files button` flickers.

## Expected behavior

The `Create support files` button should not flickers during the modal’s closing transition.

lizzy-liang-bigstack:
#### What type of PR is this?

- Fix

- Create a `MemoChildren` component to control re-rendering and prevent UI flicker in `CosModal`

https://github.com/user-attachments/assets/2a0dfd02-10b1-4902-9a66-a139046e5a93

Fixes #229 

> Note

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-ui-library/src/components/CosModal/CosModal.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosModal/CosModal.tsx
@@ -6,6 +6,7 @@ import { backdrop, modal } from './cosModalStyles'
 import { useCloseModalWithEsc } from './useCloseModalWithEsc'
 import { PropsWithClassName } from '@cube-frontend/utils'
 import { twMerge } from 'tailwind-merge'
+import MemoChildren from './MemoChildren'
 
 export type CosModalProps = {
   children: ReactNode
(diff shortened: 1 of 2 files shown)


raven-pan:
We can use `MemoChildren` to solve this problem, just like:

1. https://github.com/react-component/dialog/blob/master/src/Dialog/Content/Panel.tsx#L164
2. https://github.com/react-component/dialog/blob/master/src/Dialog/Content/MemoChildren.tsx#L7C1-L11C3

packages/cube-frontend-web-app/src/hooks/useDelayedLoading.ts
```diff
@@ -0,0 +1,49 @@
+import { useEffect, useRef, useState } from 'react'
+
+export const useDelayedLoading = (
```
steven-chiu-bigstack: `useDelayedLoading` sounds like a hook for any delayed boolean state update, but it's actually only delaying the update when closing the modal and when `isLoading` is `false`. It's difficult to come up with a good name, I think we can just put a comment at the top to briefly describe the purpose of this hook.

steven-chiu-bigstack (approved):
Hi @lizzy-liang-bigstack 

This problem is a bit tricky! Your solution works functionally, but it’s not the most concise or robust design-wise. That said, it'll take some time to get everything right (probably not worth it). Feel free to merge this PR first and handle the rest later. 🙏 

---

The `isLoading` property returned by the `useCosMutationRequest` hook indicates whether the target API is still in progress. However, in this case, we can’t use it directly because we want the create button to remain loading even after the API call succeeds, until the modal is fully hidden. To achieve this, we need a separate state that's updated manually. That's part of what `useDelayedLoading` is doing, so that's good. 👍 

We want the state to only be updated after the modal is fully hidden. Here we can use the [transitionend event](https://developer.mozilla.org/en-US/docs/Web/API/Element/transitionend_event), which fires when the CSS transition finishes. We can add an `onHidden` prop to `CosModal`, allowing users to run a custom callback once the modal is fully hidden. For example:

```tsx
const { mutateResource: createSupportFiles } = useCosMutationRequest(
  supportFilesApi.createSupportFiles
)

const [isCreating, setIsCreating] = useState(false)

const handleConfirmCreateSupportFiles = async () => {
  try {
    await createSupportFiles({ ... })
  } catch (error) {
    console.error('Create support files error: ', error)
  }
  // Note: Unlike before, we didn't update `isCreating` or clearing the modal content after the API call is finished.
}

const onModalHidden = () => {
  // Only update `isCreating` and clear modal content once the modal is fully hidden.
  setIsCreating(false)
  setSelectedNodeIds([])
}

<CosModal
  isOpen={isOpen}
  actionText="Create support files"
  actionButtonProps={{ loading: isCreating }}
  onHidden={onModalHidden}
>
  {selectedNodeIds.map(...)}
</>
```

This solution also resolves a similar/related issue where the modal content shifts or disappears immediately after clicking the close button due to a state update.

An `onShown` prop could also be added later if needed, but it's not necessary at the moment since we don't have any use cases for it.

Keep in mind that `onTransitionEnd` fires even when a child element finishes transitioning, so we need to check `e.target` to ensure the event is originated from the modal body itself.

lizzy-liang-bigstack:
Hi @steven-chiu-bigstack @raven-pan,

As discussed earlier today, this PR removes the original useDelayedLoading hook and adopts a new approach using MemoChildren to wrap the Modal content. Please feel free to let me know if there’s anything that needs to be adjusted. Thank you! 😊

PR #296, opened 2025-04-30
294 fix(web-app): node details page events table id column suffix

### LINKED ISSUE #294: [Bug]: extra "-x" is added in the event table of node detail page
## Describe the bug

as title

## To Reproduce

1. Go to 'nodes > node details'
2. Scroll down to 'event table'

## Expected behavior

we shouldn't append '-x'

## Screenshots

Image

steven-chiu-bigstack:
#### What type of PR is this?

Fix

Remove the unexpected `-${index}` suffix from the Events table in the Node Details page.

Close #294 

The `id` field in Cos Tables needs to be unique. But due to the fact that the Events API returns non-unique ids, I appended `-${index}` to each `id` to avoid duplicate key issues in React list rendering.

However, this modified `id` was also shown in the table, which confused users. To fix this, I added a new `eventId` field to hold the original event ID, and kept the modified `id` just for React rendering.

image

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeEvents.tsx
+++ b/packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeEvents.tsx
@@ -24,7 +24,11 @@ type NodeEventsProps = {
   node: Node | undefined
 }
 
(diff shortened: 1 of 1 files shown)


raven-pan (approved):
LGTM

PR #305, opened 2025-04-30
286 apply timezone to DatePicker

### LINKED ISSUE #286: [Bug] Timezone from data center needs to be reflected in the Date Picker components
1. The current Date Picker does not display the time based on the data center's timezone.
2. It might also be helpful for the UI to display the current data center timezone to reduce user confusion.

lizzy-liang-bigstack:
#### What type of PR is this?

- Feature

- Apply current timezone to the `DatePicker`.

- For example, when the timezone is set to `-10:00`, the "Today" indicator will now correctly align with the selected timezone.

Screenshot 2025-04-30 at 3 22 40 PM

- The API request time will also be aligned with the given timezone.

Screenshot 2025-04-30 at 3 21 18 PM

Fixes #286

> Note: As for the other issue mentioned in #286 — displaying the current timezone on the UI — this will be handled in a separate PR, as the design is still pending confirmation.

#### Additional documentation

```docs

```


packages/cube-frontend-web-app/src/context/CosTimeZoneProvider.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { DataCenterContext } from './DataCenterContext'
+import { CosTimeZoneContext } from '@cube-frontend/ui-library'
+
+export const CosTimeZoneProvider = (props: PropsWithChildren) => {
+  const { children } = props
+
+  const { dataCenter, isLoading: isDataCenterLoading } =
+    useContext(DataCenterContext)
+
+  if (!dataCenter || isDataCenterLoading) return null
```
raven-pan: Question: Why do we need to check `isDataCenterLoading`?

packages/cube-frontend-web-app/src/context/CosTimeZoneProvider.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+  const { utcTimeZone } = dataCenter
+
+  if (!utcTimeZone) return null
```
raven-pan: Discussion:

`utcTimeZone` is a required field in the OpenAPI spec.
Maybe we can remove this check or throw an error with message instead?

What do you think?

packages/cube-frontend-ui-library/src/components/CosTimeZone/CosTimeZoneContext.ts
```diff
@@ -0,0 +1,3 @@
+import { createContext } from 'react'
+
+export const CosTimeZoneContext = createContext<string | undefined>(undefined)
```
raven-pan: Discussion:

Maybe we can put it under the `CosDatePicker` folder?
If multiple components use `CosTimeZoneContext` in the future , we can extract it into a separate folder like `utils`?

packages/cube-frontend-ui-library/src/components/CosDatePicker/CosDatePicker.tsx
```diff
@@ -34,9 +36,17 @@ export const CosDatePicker = (props: CosDatePickerProps) => {
...
+  const timeZone = useContext(CosTimeZoneContext)
+
+  const now = useTimeZoneTime(timeZone)
+
   const [isCalendarOpen, setIsCalendarOpen] = useState(false)
 
-  const [currentMonth, setCurrentMonth] = useState(() => dayjs(new Date()))
+  const [currentMonth, setCurrentMonth] = useState(now)
+
+  const onDateClick = (date: Dayjs) => {
```
raven-pan: We can remove this function and pass `onChangeProps` directly to the context.

packages/cube-frontend-ui-library/src/components/CosDatePicker/CosDatePickerCalendar.tsx
```diff
@@ -1,59 +1,46 @@
...
+  const startDateOfCurrentMonth = currentMonth.startOf('month')
 
-  lastDate.setDate(lastDate.getDate() + (6 - lastDate.getDay()))
+  const month = startDateOfCurrentMonth.month()
 
   const weeks = useMemo(
-    () => computeCalendarWeeks(currentMonth.startOf('month')),
-    [currentMonth],
+    () => computeCalendarWeeks(startDateOfCurrentMonth),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
raven-pan: Discussion:

Maybe we can avoid using `eslint-disable-next-line`?

What if we use `startDateOfCurrentMonth` as  a memo dependency like this?

```tsx
const startDateOfCurrentMonth = useMemo(() => currentMonth.startOf('month'), [currentMonth])

const weeks = useMemo(
  () => computeCalendarWeeks(startDateOfCurrentMonth),
  [startDateOfCurrentMonth],
)
```

raven-pan (approved):

packages/cube-frontend-ui-library/src/components/CosTimeZone/CosTimeZoneContext.ts
lizzy-liang-bigstack: As of now, we know that the context will be used by the `DatePicker` and the timezone tag in the `Sidebar`. So I think it makes more sense to put the file in a separate folder. 
   
The suggestion to put the file in a `utils` folder sounds  good - but I'm wondering whether the current `CosTimeZone` folder name like `CosTimeZone` might not be ideal is because it’s not actually a component?
    
Screenshot 2025-05-02 at 10 23 00 AM

packages/cube-frontend-web-app/src/context/CosTimeZoneProvider.tsx
lizzy-liang-bigstack: I'll remove the `isDataCenterLoading` - thank you for pointing out :)

packages/cube-frontend-ui-library/src/components/CosDatePicker/CosDatePickerCalendar.tsx
lizzy-liang-bigstack: Sounds good!

PR #313, opened 2025-05-02
312 health details ng reason and error log info

### LINKED ISSUE #312: [Task]: Health Details Ng Reason and Error Log Info
1. The content of the "Reason" column should originate from `error?.reason` for ng time point.
   - Image
2. New line characters have been added to the details error log. Make them visible in the frontend.
   - Image

steven-chiu-bigstack:
#### What type of PR is this?

Feat

In the Health Details page:

1. Use `error.reason` for the "Reason" column
2. Show new line characters in details log.

Close #312 

image

### DIFF
--- a/packages/cube-frontend-api/sdk/api.ts
+++ b/packages/cube-frontend-api/sdk/api.ts
@@ -2309,6 +2309,101 @@ export interface GetIntegrationsResponseDataInner {
 /**
  * 
  * @export
+ * @interface GetLicenseAttachments500Response
+ */
+export interface GetLicenseAttachments500Response {
+    /**
+     * 
+     * @type {number}
+     * @memberof GetLicenseAttachments500Response
+     */
+    'code'?: number;
+    /**
+     * 
+     * @type {string}
+     * @memberof GetLicenseAttachments500Response
+     */
+    'msg'?: string;
+    /**
+     * 
+     * @type {string}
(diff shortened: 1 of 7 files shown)


raven-pan (approved):
LGTM!

Just a quick UI/UX discussion:

How about using `white-space-pre` and `overflow-x-scroll`? 

Screenshot 2025-05-05 at 7 26 23 PM

The UI will look like this:

https://github.com/user-attachments/assets/37a069cd-0add-49ef-91d6-53830d50a511

Current implementation:

https://github.com/user-attachments/assets/54560a70-8c98-47ef-93cc-e3e2bed12597

To me, this approach seems more readable, but the currently implementation is also fine.
Maybe we can discuss it during the next scrum meeting if we have time?

steven-chiu-bigstack:
Thanks for the great suggestion! I think using `whitespace-pre overflow-x-auto` would be better, considering that we'll have a max-width set in #325.
Without a max-width, users can resize the browser window to make the content wider, which prevents the text from wrapping into multiple lines. This can be helpful when the error log is long and has a table-like layout (similar to the console output).

However, once the max-width is applied, that table-like layout will likely wrap into multiple lines unless `overflow-x-auto` is added. I think we can just got for `whitespace-pre overflow-x-auto` and see how things go. 🙏

PR #324, opened 2025-05-05
322 feat(web-app): show months and dates in node details page charts

### LINKED ISSUE #322: [Task]: Node Details CPU and Memory Usage Charts Label Adjustments
1. Add date to the time label on hover

   Image

2. Show date with abbreviated month in the X-axis label when there's a day change.

   Image

steven-chiu-bigstack:
#### What type of PR is this?

Feat

For the charts in the Node Details page:

1. Show dates in the hover tooltip
2. Prepend `MMM DD` to the time label in the X-axis whenever there's a day change.

Close #322

Chart.js automatically hides some ticks/labels (i.e., "auto-fit") based on the available width of the chart. This means that precomputing the labels in `chartData.labels` won't always work because we can't predict which labels will end up being hidden.

To handle this, we need to wait until Chart.js has determined which labels are visible or hidden and then transform the labels accordingly. I use the [beforeFit](https://www.chartjs.org/docs/latest/api/classes/Scale.html#beforefit) lifecycle hook on the scale, which seems to work well in this case.

https://github.com/user-attachments/assets/5b482836-a8d6-4afe-979d-b9469543fb7f


raven-pan (approved):
LGTM!

PR #326, opened 2025-05-05
fix: always display node license status in copy serial number modal

### LINKED ISSUE #182: Page - License Page
[Mockup](https://www.figma.com/design/kRJ5MUAkMF0AYgXCeChtZn/COS-management-UI-Redesign-v.3?node-id=3669-22059&m=dev)

### LINKED ISSUE #307: [Bug]:  We shouldnt add extra space when copying hardware serials to clipboard
## Describe the bug

1. When click "Copy to clipboard" button in license page, it adds an extra space between hardware serials

## To Reproduce

1. Go to 'Maintenace > license'
2. Click on 'Get hardware serials'
3. Select all nodes and click 'copy to clipboard'
4. The copy text is "ToBeFilledByO.E.M., ToBeFilledByO.E.M., ToBeFilledByO.E.M." (a simply comma+space separated list)

## Expected behavior

1. The copied text should be "ToBeFilledByO.E.M.,ToBeFilledByO.E.M.,ToBeFilledByO.E.M." (a simply comma separated list without additional space)

## Screenshots

Image

raven-pan:
#### What type of PR is this?

Fix

1.) Adjusts the format of copied text when coping multiple serial numbers

Remove  the space between serial numbers, for example:
`serial A, serial B, serial C` => `serial A,serial B,serial C`.

Please see https://github.com/bigstack-oss/cube-cos-ui/issues/307.

https://github.com/user-attachments/assets/6e448600-5ace-456b-a97d-91e62fa63a46

2.) Select the `CubeCOS` product as the default when the user opens the modal.

https://github.com/user-attachments/assets/b65ffd12-455e-4044-bb58-c66b3e99da5c

3.) Always displays the license status for every host.

If a host has no license, it will display `Unlicense`. 
This requires integration with the [new COS License Attachment API](https://github.com/bigstack-oss/cube-cos-openapi/pull/35).

https://github.com/user-attachments/assets/b4a74f60-68a3-4bff-8df7-1cb7561e3123

## Note

1. One host can have multiple licenses, with 0 or 1 licenses per product.
2. One license can affect multiple hosts.

Close https://github.com/bigstack-oss/cube-cos-ui/issues/307, Close #182


packages/cube-frontend-web-app/src/pages/maintenance/license/_components/LicenseActions/HardwareSerialNumberModal/useCopySerialNumber.ts
```diff
@@ -1,17 +1,21 @@
-import { Node } from '@cube-frontend/api'
-import { useState, useRef, useEffect } from 'react'
+import { useEffect, useRef, useState } from 'react'
+import { BatchLicenseAttachmentTableRow } from './LicenseAttachmentTable'
 
 const MESSAGE_TIMEOUT = 2 * 1000
 
-export const useCopySerialNumber = (selectedNodes: Node[]) => {
+export const useCopySerialNumber = (
```
steven-chiu-bigstack, line 6: Spotted this by accident: Is it correct that the `clearTimeout` in this hook’s effect should be wrapped inside a return statement?:

{4FD7D3E5-93B1-4F92-BFFD-BC65998C54D6}

packages/cube-frontend-web-app/src/pages/maintenance/license/_components/LicenseActions/HardwareSerialNumberModal/useLicenseAttachmentRowSelection.ts
```diff
@@ -0,0 +1,52 @@
...
+        ) || []
+      setSelectedRowIds(allRowIds)
+    } else {
+      setSelectedRowIds([])
+    }
+  }
+
+  useEffect(() => {
+    setSelectedRowIds([])
+  }, [licenseAttachmentRows])
```
steven-chiu-bigstack, line 40: This effect is duplicated on line 17.

{77462A4E-6B96-46AD-B708-8A6E01EFB855}

packages/cube-frontend-web-app/src/pages/maintenance/license/_components/LicenseActions/HardwareSerialNumberModal/useLicenseAttachmentTable.ts
```diff
@@ -0,0 +1,87 @@
...
+    setSearchKeyword('')
+    setDebounceSearchKeyword('')
+  }
+
+  const clearFilters = () => {
+    setSelectedProduct(undefined)
+    setSearchKeyword('')
+    setSelectedRoles([])
+    setSelectedNodeLicenseStatuses([])
+    onCloseClick()
```
steven-chiu-bigstack: A couple of questions regarding this hook:

1. Calling `onCloseClick` here feels a bit strange and possibly unnecessary. Do you think it makes sense to remove it?
2. Since CubeCOS is the default selected product and there's no way to clear the selection, could we assign a default value to `selectedProduct` when declaring it?
   - For example: `const [selectedProduct, setSelectedProduct] = useState(GetNodesProductsEnum.CubeCos)`
   - This way, `selectedProduct` will always be truthy, allowing us to remove the `selectedProduct && 'h-[400px]'` check in `HardwareSerialNumberModal` (always has `h-[400px]`).
   - That means the null check for `selectedProduct` in `useCosGetRequest` could also be removed. It feels fine to fetch license attachments when entering this page, even if this modal never gets opened. But if it's slow or expensive we can still introduce the `isOpen` state as a prop and prevent the request from firing when the modal is closed.
3. We could also rename the `clearFilters` function to something like `resetFilters`, and reset `selectedProduct` back to `GetNodesProductsEnum.CubeCos` instead of assigning `undefined`.
   - That would let us remove both the props dependency and the `isOpen` effect below from this hook, since this function is called by the parent modal before closing.

steven-chiu-bigstack (approved):
Looks good, nice work @raven-pan! 🙏 
I left a few minor comments for you to check. Feel free to merge the PR once they're fixed, thanks!

packages/cube-frontend-web-app/src/pages/maintenance/license/_components/LicenseActions/HardwareSerialNumberModal/useCopySerialNumber.ts
raven-pan: Good catch!
Thanks for you reminder.

packages/cube-frontend-web-app/src/pages/maintenance/license/_components/LicenseActions/HardwareSerialNumberModal/useLicenseAttachmentTable.ts
raven-pan: Thanks for you advice, this will be much better.

PR #334, opened 2025-05-06
docs(rpm): make the commands work for reinstallations

Eandalf-Bigstack:
#### What type of PR is this?

- documentation

- List out the working commands for re-installations

- Commands for re-installations do not work


steven-chiu-bigstack (approved):
-y ! 👍

PR #351, opened 2025-05-08
fix: 113 293 svg and canvas chart break page RWD

### LINKED ISSUE #113: UI Library - CosSegmentedBar Break Flexbox After Resize Browser Window
Screen Recording 2025-02-24 at 6 22 21 PM

### LINKED ISSUE #293: [Task] COS panel RWD
We might need considering RWD for home page panel.
https://github.com/user-attachments/assets/9e2a9384-7b8a-427a-8182-d16d0309687d

1. Top: Pie Chart move down below when the width become narrow.
2. Health can scroll.
3. Event & Nodes are OK. (They can scroll)

1. All can be scroll
2. Fix the panel
3. Resize

**Additional context**

raven-pan:
#### What type of PR is this?

Fix

Fixes #113, Fixes #293

1.) Fixes Home Page responsiveness (RWD)

https://github.com/user-attachments/assets/43caf3cf-72a0-4740-bcba-a0cc758d1a77

2.) Fixes Chart Page responsiveness (RWD)

https://github.com/user-attachments/assets/99916c85-39d5-4b9e-9c81-10a9fa9269dc

#### Notes

1. When using the [Chart.js](https://www.chartjs.org/docs/latest/) component, you need to pass `responsive: true` to the chart options and wrap it with a `relative` DOM container. See the [Chart.js Responsive Doc](https://www.chartjs.org/docs/latest/configuration/responsive.html) for proper responsive setup.
2. If you're using any `svg` or `canvas` chart inside a flex item, you need to set a `min-width` to make `flex-shrink` work properly. This is because the chart component calculates the size of `svg` or `canvas` size based on the width of their parent DOM using Javascript. Please related discussion: https://github.com/chartjs/Chart.js/issues/4156#issuecomment-295180128 and https://jsfiddle.net/pgg9pz7c/3/.
3. I added a `useMediaQuery` hook for ChartPanel responsive display, since it's not possible to move the Pie Chart down correctly on narrow screen using CSS `flex-wrap`.
4. I added a `ScrollContainer` component to handle scroll behavior and to fix the issue where the child box-shadow was being cut off by the parent., Please see [this discussion](https://stackoverflow.com/questions/70802682/parents-overflowhidden-doesnt-show-childs-box-shadow) for more details.


lizzy-liang-bigstack (approved):
@raven-pan Thank you for the update and the cool solution to the svg and canvas resizing issue within a flexbox layout — I learned a lot from the references you shared in the PR!

That said, I did notice that the `Storage` panel on the `Home/Chart` page causes unexpectedly overflowing on smaller screens. I'm not sure if this is something we need to address in phase 1, since we don’t yet have a complete responsive design plan, and the situation shown in the video appears to be an edge case.

Feel free to share your thoughts on whether we should fix this now or leave it for a future phase once the RWD design is more defined. 

https://github.com/user-attachments/assets/b8d05dd5-6263-4a03-a314-7985e1c2213b

raven-pan:
> @raven-pan Thank you for the update and the cool solution to the svg and canvas resizing issue within a flexbox layout — I learned a lot from the references you shared in the PR!
> 
> That said, I did notice that the `Storage` panel on the `Home/Chart` page causes unexpectedly overflowing on smaller screens. I'm not sure if this is something we need to address in phase 1, since we don’t yet have a complete responsive design plan, and the situation shown in the video appears to be an edge case.
> 
> Feel free to share your thoughts on whether we should fix this now or leave it for a future phase once the RWD design is more defined.
> 
>  Screen.Recording.2025-05-09.at.4.10.47.PM.mov

@lizzy-liang-bigstack  Good catch!

I fixed this issue in this [commit](https://github.com/bigstack-oss/cube-cos-ui/pull/351/commits/544d1bb6d5bed7730dff3855c969414db1f1fe24).

The horizontal scrollbar was caused by the `Dropdown` component, which includes a floating absolute positioned DOM element.

https://github.com/user-attachments/assets/8cc53f9a-8f14-43c6-9e53-0bf48ed214e7

The solution is to add `overflow-x: hidden` style to the body element.

References:

1. https://css-tricks.com/findingfixing-unintended-body-overflow/
2. https://www.reddit.com/r/webdev/comments/1ciwixx/preventing_overflow_with_absolutepositioned/?rdt=39712

PR #357, opened 2025-05-09
348: fix(web-app): support files page freeze when attempting to download files that are not ready

### LINKED ISSUE #348: [Bug]: Support Files Page Freeze When Attempting to Download Files That Are Not Ready
## Describe the bug

Users reported that the screen freezes when attempting to download support files that are not ready.

Image

Disable the Download button or find a way to inform users that the support files is still being created based on the support file status:

Image

steven-chiu-bigstack:
#### What type of PR is this?

Fix

Users reported that attempting to download support files that are not yet ready will cause the page to freeze. Hence, the Download button is replaced by a loading spinner and a "Creating..." text when the files are still being created.

Close #348 

The designers have agreed on using this solution.

https://github.com/user-attachments/assets/7a0b5d63-7343-4a71-93e6-27b4ec6e6e4c


packages/cube-frontend-web-app/src/pages/home/overview/_components/NodePanel/NodePanel.tsx
```diff
@@ -37,6 +37,15 @@ export const NodePanel = () => {
...
   const updateTime = useUpdateTime(nodesData, isLoading)
 
+  const rows = useMemo<Node[]>(() => {
+    const nodes = nodesData?.nodes ?? []
+    return nodes.map((node) => ({
+      ...node,
+      // Adjust `id` because `id` will be an empty string when the node is in `down` status.
+      id: node.id || uniqueId('node'),
+    }))
+  }, [nodesData?.nodes])
```
steven-chiu-bigstack, line 47: I sneaked in some changes regarding the node table found this morning. The `id` (as long as some other fields such as `address`, `serialNumber`, etc.) of the nodes returned by get nodes API will be an empty string if the node is in `down` status. This will mess with the table rendering when there are more than 2 down nodes because the `id` field is used as the key of the table rows during rendering.

So I use the `uniqueId()` function from lodash to ensure there's always a unique value for `id`.

raven-pan (approved):
LGTM!

raven-pan (approved):
Just a minor thought: 

If I were the user, I would expect the LoadingSpinner to eventually become  a download link if I wait long enough on the same page.

However, this would require real-time support from both the UI and API.
I think it's too much effort for Phase 1.

steven-chiu-bigstack:
> Just a minor thought: 
> 
> 
> 
> If I were the user, I would expect the LoadingSpinner to eventually become  a download link if I wait long enough on the same page.
> 
> 
> 
> However, this would require real-time support from both the UI and API.
> 
> I think it's too much effort for Phase 1.

Good point! I thought that was already achieved with the short polling, but now that you mention it, I noticed we didn't have that on the Support Files page.

Do you think adding `useInterval` to this page to re-fetch data every few seconds would cause any problem?

steven-chiu-bigstack:
As discussed on Slack, polling is inappropriate because of the filter and pagination functionalities on this page.
We can think about a better solution in phase 2, thanks!

PR #359, opened 2025-05-09
346 conditionally put `password`  in the patch request

### LINKED ISSUE #346: [TASK] Settings (mail sender) - Not put the `password` key in the email sender patch api if user doesn't edit the password

Currently, we realized that we can't identify whether user is requesting password change or not because we allow password can be set in empty, 

Image

also, the current password string won't be brought to UI, so when user call patch api, we're not sure the actual intention is to "set empty password" or "password is not changed"

<br/>
<br/>

As discussed with @steven-chiu-bigstack , we decide to make the behavior to be

- no password changes: to not put the `password` key in the email sender patch api

- has password changes: put the `password` key in the email sender patch api

<br/>

lizzy-liang-bigstack:
#### What type of PR is this?

- Feat

- If a new password is provided, include it in the patch request payload.

https://github.com/user-attachments/assets/12278368-718e-4818-bf75-153edd473cd9

- If no password is provided, omit the `password` key.

https://github.com/user-attachments/assets/49f0210f-3198-43af-b9a2-5cabab461ed9

Fixes #346 

> Note

#### Additional documentation

```docs

```

https://github.com/user-attachments/assets/5120c513-f48a-49bd-aaf1-03dd5a071cb9

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/settings/_components/email/senders/emailSenderMappers.ts
+++ b/packages/cube-frontend-web-app/src/pages/settings/_components/email/senders/emailSenderMappers.ts
@@ -37,10 +37,22 @@ export const rowToEmailSenderPostRequest = (
 
 export const rowToEmailSenderPatchRequest = (
   row: EmailSenderRow,
-): EmailSenderPatchRequest => ({
-  from: row.from,
-  host: row.host,
-  port: parseInt(row.port),
-  username: row.username,
-  password: row.password,
-})
+): EmailSenderPatchRequest => {
+  const request: EmailSenderPatchRequest = {
+    from: row.from,
+    host: row.host,
+    port: parseInt(row.port),
+    username: row.username,
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


packages/cube-frontend-web-app/src/pages/settings/_components/email/senders/emailSenderMappers.ts
```diff
@@ -37,10 +37,22 @@ export const rowToEmailSenderPostRequest = (
...
+  }
+
+  /**
+   * Conditionally put the `password` field in the patch request:
+   * - If no password is provided, omit the `password` key to avoid unnecessary updates.
+   * - If a new password is provided, include it in the patch request payload.
+   */
+  if (row.password) {
+    request.password = row.password
+  }
```
steven-chiu-bigstack, line 55: The current implementation doesn't cover the case of updating the password from a truthy value to an empty string, as empty strings are treated as falsy in JS.

PR #363, opened 2025-05-12
130: fix(ui-library): cos progress bar border radius

### LINKED ISSUE #130: UI Library - Progress Bar Incorrect display for 1% and 99%
Border incorrect display:

Image

steven-chiu-bigstack:
#### What type of PR is this?

Fix

Fix the border-radius issue for `CosProgressBar`.

Close #130 

This problem occurs not only when the filled percentage is 1% and 99%. The root cause is that if the bar width is too small, the filled area will be too short for the border-radius to appear (e.g. when `progress` = 1%). Likewise, if the filled area is too long, the remaining blank space will be too narrow for the trailing border-radius to show (e.g., when `progress` = 99%). It depends on the width of the progress bar.

{29959E73-5518-40AB-89D2-2D36BCEFE2A0}

Similar issue also occurs with SVG. While SVG does solve the border-radius problem, the filled area can still be too short to be visible, making it an imperfect solution:

image

In theory, we should also adjust `CosSegmentedBar` to prevent the same problem from happening; luckily, `CosSegmentedBar` is only used in much wider areas due to the extra info it contains, so I decided to leave it as is for now.


raven-pan (approved):
LGTM!

PR #372, opened 2025-05-13
fix: 366 inconsistent name case format for OpenStack

### LINKED ISSUE #366: [Bug]: Inconsistent name case format for openstack
## Describe the bug

We should use OpenStack where o and s are uppercase for displaying openstack

## Version

v3.0.0

## To Reproduce

1. Go to 'Integrations'
2. Check OpenStack and it use "Openstack"

## Expected behavior

We should use OpenStack

## Screenshots

Image

##Additional context

raven-pan:
#### What type of PR is this?

Fix

Fixes #366 

Issue: Please see Fixes #366

Solution: 

Extracted `integrationUIData` into `utils/integration.ts`, so both `CosHeader` and the `Integration Page` can share the same icon and display logic.

Screenshot 2025-05-13 at 4 13 05 PM


steven-chiu-bigstack (approved):
LGTM!

PR #381, opened 2025-05-13
353 fix(web-app): rename COS license expiration table header back to license expiration

### LINKED ISSUE #353: [Task] Rename License Expiration to COS License Expiration in the Node Table and Node Details
Since one host can have multiple licenses, with 0 or 1 license per product, we need a UI solution to display multiple license details.

In Phase 1, we will display the most important license—COS License—and rename the label from License Expiration to COS License Expiration.

A more comprehensive UI solution can be discussed and implemented in Phase 2.

Image

Image

Image

cc
@jdjgya @steven-chiu-bigstack @giselech

steven-chiu-bigstack:
#### What type of PR is this?

Fix

Rename "COS License Expiration" table headers back to "License Expiration".

Close #353

I think that makes sense. As the original issue description has outlined:

> Since one host can have multiple licenses, with 0 or 1 license per product...

If we just show "COS License Expiration", users may get confused if the host has only CubeCMP license(s) installed

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/home/overview/_components/NodePanel/NodeTable.tsx
+++ b/packages/cube-frontend-web-app/src/pages/home/overview/_components/NodePanel/NodeTable.tsx
@@ -65,7 +65,7 @@ export const NodeTable = (props: NodeTableProps) => {
           </CosTag>
         )}
       </BasicNodeTable.Column>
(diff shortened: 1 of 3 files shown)


lizzy-liang-bigstack (approved):
LGTM

PR #418, opened 2025-05-15
415 license page and health module value casing

### LINKED ISSUE #415: [Bug]: Some of strings have inconsistent upper/lower case
## Describe the bug

Some of strings have inconsistent upper/lower case

<br/>

## Version

COS 3.0.0 20250513-2025

<br/>

## To Reproduce

Env: 10.32.10.150 (3cc)

<br/>

<br/>

1). Log in to 10.32.10.150 (3cc)

<br/>

2). Go to License page, and see the `feature` is `N/a`, but api returns `N/A`

Quantity and Support Plan also using `N/A`, so not sure if `feature` should align those convention?

Image

<br/>

3). Go to License page > Import license, then import an old license, then will see the same symptom of (2)

Image

<br/>

4). Go to health list page, and saw the first char of `etcd` and `nginx` is lowercase, seems like should use the uppercase for that?

Image

Image

<br/>

## Expected behavior

<br/>

steven-chiu-bigstack:
#### What type of PR is this?

Fix

Fix UI texts casing.

Close #415 

Issues with etcd were fixed yesterday, but nginx was missing from the known module list. We use a map to convert module names to labels because there are non-intuitive cases (e.g., "iaasDb" -> "IaaS DB", "k8saas" -> "K8SaaS"), and the service/module names that's not in the map will be displayed as-is.

Image

I’ve added "nginx" to the map and updated the default behavior to capitalize the first letter of unmapped names, which should cover most scenarios.

Image

https://github.com/user-attachments/assets/cbf7d246-1e4e-41a2-93e2-c54ab434e00b

#### Additional documentation

Lodash's [capitalize](https://lodash.com/docs/4.17.15#capitalize) function converts first character of a string to upper case and **converts the remaining to lower case**. That means `capitalize('HeLLo')` outputs `Hello`. This is so counterintuitive. 😢😢😢 We should use [upperFirst](https://lodash.com/docs/4.17.15#upperFirst) instead.

### DIFF
--- a/packages/cube-frontend-ui-library/src/stories/components/CosTables/ViewDetails/CosViewDetailsTable.stories.tsx
+++ b/packages/cube-frontend-ui-library/src/stories/components/CosTables/ViewDetails/CosViewDetailsTable.stories.tsx
@@ -3,7 +3,7 @@ import {
   useExpandedRowIdSet,
 } from '@cube-frontend/ui-library'
 import { Meta, StoryObj } from '@storybook/react'
-import { capitalize } from 'lodash'
+import { upperFirst } from 'lodash'
 import { StoryLayout } from '../../../../internal/components/StoryLayout/StoryLayout'
 import {
   isCmpLicense,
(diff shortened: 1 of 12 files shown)


lizzy-liang-bigstack (approved):
LGTM!

PR #421, opened 2025-05-15
chore: fix dependabot

bigstack-brian-su:
- add schedule scanning to monday 9am
- group pull request to reduce fatigue
- set pull request sensitivity

#### What type of PR is this?

Fix

Fixes #

> Note

#### Additional documentation

```docs

```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,6 +20,7 @@ updates:
       day: monday
       time: "09:00"
       timezone: "Asia/Taipei"
+    open-pull-requests-limit: 0
     groups:
       minor-and-patch:
         applies-to: version-updates
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,9 +4,8 @@ updates:
     directory: /
     schedule:
       interval: weekly
-      day: monday
-      time: 09:00
-
+      time: "09:00"
+      timezone: "Asia/Taipei"
   - package-ecosystem: npm
     directories:
       - /
@@ -19,8 +18,8 @@ updates:
     schedule:
       interval: weekly
       day: monday
-      time: 09:00
-    update-types: ["version-update:semver-patch"]
+      time: "09:00"
+      timezone: "Asia/Taipei"
     groups:
       minor-and-patch:
         applies-to: version-updates
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,44 +3,35 @@ updates:
   - package-ecosystem: github-actions
     directory: /
     schedule:
-      interval: daily
+      interval: weekly
+      day: monday
+      time: 09:00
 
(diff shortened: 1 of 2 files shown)


steven-chiu-bigstack (approved):
Hi @bigstack-brian-su 

The format and configuration look good based on the [Dependabot docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference), thanks for taking care of this! 🙏 

Since we don't have unit tests and E2E tests at the moment, dependency updates could potentially break functionality without anyone knowing it, though this is less likely for minor and patch updates. However, it's not quite possible to implement E2E tests for all pages considering the time it takes, so it's probably better to do necessary updates (e.g., security updates) as early as possible before the application grows even larger.

For the `security-updates` group, I think it makes sense to always update those packages as recommended, as they typically don’t affect functionality or change library interfaces.

But I'm not quite sure what is the right thing to do with the `version-updates` group. We could either go with the current setup and close the version-updates PR every week or review the changelogs for each package and upgrade only if there's no major conerns.

Simply put, in my opinion, the simplest solution for us at the current stage might be:

1. Ignore `version-updates` by closing the weekly PR or commenting them out in this config file until E2E coverage improves, and
2. Apply all `security-updates`.

Does that make sense? 🤔 @bigstack-brian-su  @raven-pan @lizzy-liang-bigstack

bigstack-brian-su:
> Hi @bigstack-brian-su
> 
> The format and configuration look good based on the [Dependabot docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference), thanks for taking care of this! 🙏
> 
> Since we don't have unit tests and E2E tests at the moment, dependency updates could potentially break functionality without anyone knowing it, though this is less likely for minor and patch updates. However, it's not quite possible to implement E2E tests for all pages considering the time it takes, so it's probably better to do necessary updates (e.g., security updates) as early as possible before the application grows even larger.
> 
> For the `security-updates` group, I think it makes sense to always update those packages as recommended, as they typically don’t affect functionality or change library interfaces.
> 
> But I'm not quite sure what is the right thing to do with the `version-updates` group. We could either go with the current setup and close the version-updates PR every week or review the changelogs for each package and upgrade only if there's no major conerns.
> 
> Simply put, in my opinion, the simplest solution for us at the current stage might be:
> 
> 1. Ignore `version-updates` by closing the weekly PR or commenting them out in this config file until E2E coverage improves, and
> 2. Apply all `security-updates`.
> 
> Does that make sense? 🤔 @bigstack-brian-su @raven-pan @lizzy-liang-bigstack

Yes, let me adjust it to not open any pull requests for `version-updates` but it will still generate the alerts.

abemscac:
Thanks @bigstack-brian-su! I think we can merge this PR and see how it goes. 🙏

PR #434, opened 2025-05-19
426 use cos mutation request data less requests

### LINKED ISSUE #426: [Refactor]: Add Support for Data-Less Requests in the `useCosMutationRequest` Hook
Some mutation APIs returns no `data` in the response, but the current implementation of the `useCosMutationRequest` expects `data` to be defined in the response interface. For now we use the `as` type casting as a workaround to this problem (e.g., `triggers/[create]/useUpdateTrigger.ts:43`):

- With Type Casting
   Image
- Without Type Casting (TS error shows up)
   Image

Adjust the `useCosMutationRequest` so that we can remove all similar `as` type castings from the codebase.

steven-chiu-bigstack:
#### What type of PR is this?

Refactor

Add support for data-less requests in the `useCosMutationRequest` hook.

Close #426 

- Type castings have been removed.
   - image
- The `data` field for data-less mutation requests will be of the `unknown` type.
   - We could've made it `never`, but that would require less-readable code and the trade-off probably isn’t worth it.
   - image


raven-pan (approved):
LGTM!

PR #439, opened 2025-05-19
fix: 433 update the license nagging after a new license file imported

### LINKED ISSUE #433: [Enhancement] License nagging message - Do GET datacenter api immediately after license is being imported
## Describe the bug

@chase-chou @Bigstack-Nate both mentioned that the UI and Backend should update the license data / data center data right away to have better experience after license is being imported.

Backend already enhanced such problem in the RC1, but realized UI seems like won't do `GET datacenter api` right away or periodically for nagging msg update, so create a ticket to FE for enhancement

<br/>

## Version

COS 3.0.0 RC1

<br/>

## To Reproduce

In the video below, we can see that we have to F5(refresh) the page to get the newest data from API.

https://github.com/user-attachments/assets/56b82fd2-eeae-4799-a4c6-a117d9a47d0f

raven-pan:
#### What type of PR is this?

Fix

Fixes #433

Problem:
The License Nagging should be updated after a license file is imported to reflect the latest license status of the data center. 

Solution:
Re-fetch `dataCenters` after successfully importing the license file .

#### Test Screenshots

https://github.com/user-attachments/assets/3db7b74a-55df-4055-b496-3cc338965c1a

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/maintenance/license/MaintenanceLicensePage.tsx
+++ b/packages/cube-frontend-web-app/src/pages/maintenance/license/MaintenanceLicensePage.tsx
@@ -21,7 +21,7 @@ import { useTopLicenseNaggingStore } from '@cube-frontend/web-app/stores/topLice
 import { LicenseActions } from './_components/LicenseActions/LicenseActions'
 
 export const MaintenanceLicensePage = () => {
-  const { dataCenter } = useContext(DataCenterContext)
+  const { dataCenter, fetchDataCenters } = useContext(DataCenterContext)
 
   const [searchKeyword, setSearchKeyword] = useState<string>('')
   const [selectedProducts, setSelectedProducts] = useState<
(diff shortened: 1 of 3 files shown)


lizzy-liang-bigstack (approved):
LGTM - thank you for the update! 🚀

packages/cube-frontend-web-app/src/pages/maintenance/license/MaintenanceLicensePage.tsx
```diff
@@ -52,6 +52,11 @@ export const MaintenanceLicensePage = () => {
     } satisfies LicensesApiGetLicensesRequest
   })
 
+  const handleLicenseImportSuccess = () => {
+    fetchLicenses()
+    fetchDataCenters!()
```
steven-chiu-bigstack, line 57: Nitpick: Logically, `fetchLicenses` should be called after `fetchDataCenters` is done because it needs the data center name. But data center name is unlikely to change (at least at the current stage), and we can only get the updated data center in the next render, which could overcomplicate things because we'll need a `useEffect` to call `fetchLicenses` in that case.

In short, I think the current implementation looks good!👌 Just have to keep this in mind if data center names ever become mutable in the future.

PR #456, opened 2025-05-20
fix: set start and end of day when selected dates are the same

lizzy-liang-bigstack:
#### What type of PR is this?

- Fix

- When the user selects the same day with the `DatePicker`, it will set the start and end times to the beginning and end of that day.

Screenshot 2025-05-20 at 3 05 37 PM

Fixes https://github.com/bigstack-oss/cubecos/issues/84

> Note

#### Additional documentation

```docs

```


steven-chiu-bigstack (approved):
LGTM!

raven-pan (approved):
This was my issue.
I forget to use `startOfDay` and `endOfDay` to get the correct day range for the request.
Thank you for fixing it.

PR #463, opened 2025-05-21
Fix/448 node support files and license page should return to page 1 after filters changed

### LINKED ISSUE #448: [Bug]:  Node, Support Files and License Page should back to page 1 when filter conditions or itemsPerPage changed.
## Describe the bug

The Node, Support Files and License Page should back to page 1 when filter conditions or `itemsPerPage` change, but they do not.

## Version

3.0.0-rc1

## To Reproduce

1. Go to the Support Files page.
2. Navigate to page 2.
3. Change any filter.

## Actual behavior

The table should return to page 1.
The table remains on page 2.

## Screenshots

https://github.com/user-attachments/assets/b046acf3-dcc1-4917-8bd0-8b0d67e6f252

raven-pan:
#### What type of PR is this?

Fix

Fixes #448

Fixes:

1. The Node, Support Files and License page should retain filter conditions after a page refresh by using URLSearchParams.
2. The Node, Support Files and License page should return to page 1 when filters are changed.

Solution:

1. Extract `useSearchParamsQuery` hook to sync request query and URLSearchParams.
3. Define request query schemas using Zod.
4. Infer  request query types from the Zod schema.

#### Screenshots

1.) The Node List Page

https://github.com/user-attachments/assets/820525c4-e54f-4b11-8ad7-eeb736a249b5

2.) The Support Files List Page

https://github.com/user-attachments/assets/a446117b-19c0-4589-af1f-1ad8b280e2bd

3.) The License List Page

https://github.com/user-attachments/assets/ea864726-dd0c-41cf-a104-11f67e1928bf

4.) The Tuning List Page

https://github.com/user-attachments/assets/11b1dbb5-bbc5-436c-8734-967067f92532

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/maintenance/license/MaintenanceLicensePage.tsx
+++ b/packages/cube-frontend-web-app/src/pages/maintenance/license/MaintenanceLicensePage.tsx
@@ -1,15 +1,9 @@
-import { useContext, useEffect, useState } from 'react'
-import {
-  GetLicensesProductsEnum,
-  ListLicenseCurrentStatus,
-  GetLicensesTypesEnum,
-  LicensesApiGetLicensesRequest,
-} from '@cube-frontend/api'
+import { useContext, useEffect } from 'react'
+import { LicensesApiGetLicensesRequest } from '@cube-frontend/api'
 import {
   CosGeneralPanel,
   CosPagination,
   CosStroke,
-  DEFAULT_ITEMS_PER_PAGE,
 } from '@cube-frontend/ui-library'
 import { licenseApi } from '@cube-frontend/web-app/api/cosApi'
 import { DataCenterContext } from '@cube-frontend/web-app/context/DataCenterContext'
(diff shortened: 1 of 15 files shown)


packages/cube-frontend-web-app/src/pages/maintenance/license/MaintenanceLicensePage.tsx
```diff
@@ -19,22 +13,29 @@ import { useDebounce } from '@cube-frontend/web-app/hooks/useDebounce'
...
+    onItemsPerPageChange,
+  } = useLicenseListQuery()
+
+  const [debouncedSearchKeyword, setDebounceSearchKeyword] = useDebounce(
+    query.keyword,
+    300,
+  )
+  const handleSearchKeywordClear = () => {
+    onKeywordChange('')
+    setDebounceSearchKeyword('')
```
lizzy-liang-bigstack, line 37: Discussion:
Since `debouncedSearchKeyword` comes from `query.keyword`, and calling `onKeywordChange('')` should update it accordingly, do you think we could remove the `setDebouncedSearchKeyword('')` here?

Or is it intended to clear `debouncedSearchKeyword` immediately for a better UX? If that’s the case, I think the current approach makes sense.

packages/cube-frontend-web-app/src/pages/node/NodeListPage.tsx
```diff
@@ -107,18 +107,18 @@ export const NodeListPage = () => {
           <NodeTable
             rows={rows}
             isLoading={isLoading}
-            skeletonRowCount={pageSize}
+            skeletonRowCount={query.currentPage}
```
lizzy-liang-bigstack: Maybe `query.itemsPerPage` would be more suitable in this case?

packages/cube-frontend-web-app/src/utils/pagination.ts
```diff
@@ -0,0 +1,22 @@
+import { z } from 'zod'
+import { DEFAULT_ITEMS_PER_PAGE, ItemsPerPage } from '@cube-frontend/ui-library'
+
+export const paginationQuerySchema = z.object({
+  currentPage: z
+    .string()
+    .nullable()
+    .transform((value) => {
+      return value ? parseInt(value, 10) : 1
+    }),
```
lizzy-liang-bigstack, line 10: Discussion:
To set a default value in the schema here, I wonder if we could consider using `.default(...)` provided by Zod. It might make the intention clearer and reduce the need for additional checks later. What do you think?

Reference: https://v3.zod.dev/?id=default

lizzy-liang-bigstack (approved):
Hi @raven-pan, thanks for the update! I've left some minor comments - the rest seems perfectly fine with me!

packages/cube-frontend-web-app/src/pages/maintenance/license/_components/utils.ts
```diff
@@ -99,3 +105,113 @@ export const mapToTableRows = (
...
+  ItemsPerPage = 'pageSize',
+}
+
+const licenseListQuerySchema = paginationQuerySchema.extend({
+  keyword: z
+    .string()
+    .nullable()
+    .transform((value) => {
+      return value ?? ''
+    }),
```
lizzy-liang-bigstack, line 124: Discussion:
To set a default value in the schema here, I wonder if we could consider using `.default(...)` provided by Zod. It might make the intention clearer and reduce the need for additional checks later. What do you think?

Reference: https://v3.zod.dev/?id=default

packages/cube-frontend-web-app/src/pages/maintenance/license/MaintenanceLicensePage.tsx
raven-pan: This is intentional for better UX.

packages/cube-frontend-web-app/src/pages/node/NodeListPage.tsx
raven-pan: Good catch!
I'll fix it.

PR #471, opened 2025-05-22
fix(web-app): tunings table state sync bug

steven-chiu-bigstack:
#### What type of PR is this?

Fix

Fix tunings table updating status sync bug.

Close https://github.com/bigstack-oss/cubecos/issues/102

https://github.com/user-attachments/assets/297e3e34-c485-4ecf-bbae-91893ed2a3f7

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/events/tunings/useTuningRows.ts
+++ b/packages/cube-frontend-web-app/src/pages/events/tunings/useTuningRows.ts
@@ -2,9 +2,10 @@ import { Page, TuningsApiListTuningsRequest } from '@cube-frontend/api'
 import { tuningsApi } from '@cube-frontend/web-app/api/cosApi'
 import { DataCenterContext } from '@cube-frontend/web-app/context/DataCenterContext'
 import { useCosGetRequest } from '@cube-frontend/web-app/hooks/useCosRequest/useCosGetRequest'
-import { useInterval } from '@cube-frontend/web-app/hooks/useInterval'
+import { useSequentialInterval } from '@cube-frontend/web-app/hooks/useSequentialInterval/useSequentialInterval'
 import { parseErrorMessage } from '@cube-frontend/web-app/utils/errorMessage'
-import { useContext, useEffect, useMemo, useState } from 'react'
+import { isEqual } from 'lodash'
+import { useContext, useEffect, useMemo, useRef, useState } from 'react'
 import { TuningRow, tuningToRow } from './tuningsUtils'
 import { ListTuningsQuery } from './useListTuningsQuery'
 
@@ -25,6 +26,8 @@ export const useTuningRows = (
 
   const [rows, setRows] = useState<TuningRow[]>([])
 
+  const intervenedRowIdsRef = useRef<Set<string>>(new Set())
+
   const {
     data: listTuningsResponse,
     hasResponseBeenReceived,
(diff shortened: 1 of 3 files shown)


packages/cube-frontend-web-app/src/pages/events/tunings/useTuningRows.ts
```diff
@@ -25,6 +27,8 @@ export const useTuningRows = (
 
   const [rows, setRows] = useState<TuningRow[]>([])
 
+  const intervenedRowIdsRef = useRef<Set<string>>(new Set())
```
steven-chiu-bigstack, line 29: For the case where polling request fired -> user clicks on the toggle (and pause the polling) -> previous polling response arrived, all table rows will still be overwritten with the API response.

Thus, we need to keep track of the rows that should not be overridden and manually update them after the toggle/reset action is done.

packages/cube-frontend-web-app/src/pages/events/tunings/useTuningRows.ts
```diff
@@ -149,6 +176,43 @@ export const useTuningRows = (
...
+      // The combination of tuning name and hosts is guaranteed to be unique.
+      // However, the API returns tunings matching any host in the `host`
+      // parameter, so we have to compare the host names array to find the
+      // matching tuning.
+      const response = await tuningsApi.listTunings({
+        dataCenter: dataCenter!.name,
+        host: row.hosts.map((host) => host.name),
+        keyword: row.name,
+        pageNum: 1,
+        pageSize: DEFAULT_ITEMS_PER_PAGE,
```
raven-pan: Discussion:
Maybe in some cases, pageNum: 1 and pageSize: 10 won’t include the desired tuning row?

packages/cube-frontend-web-app/src/pages/events/tunings/useTuningRows.ts
```diff
@@ -129,16 +154,18 @@ export const useTuningRows = (
     })
```
raven-pan, line 161: Discussion:
Should we add the row id to intervenedRowIdsRef after calling `resetTuning`?

raven-pan (approved):

packages/cube-frontend-web-app/src/pages/events/tunings/useTuningRows.ts
steven-chiu-bigstack: That's correct ,good catch!

packages/cube-frontend-web-app/src/pages/events/tunings/useTuningRows.ts
steven-chiu-bigstack: That's strange. 🤔 
I wonder what's the API response when that happened? Did the row get "split in half"? (one for 451 and the other for 452&453)
Related code in the backend has been updated & deployed manually during my call with Bjergsen this afternoon, and I heard him say he might make some adjustments to the API if applicable. I'll confirm this with him tomorrow. 🙏

jdjgya: 我聞到了呼喚我的味道 (聞聞 痾不是

<br/>

in the current control flow, if a tuning parameter has totally same value in the 451 452 453 , the it should keep only one row with these 3 hosts in the case of enable, disable, and reset. 

<br/>

if an user set a new value to 451 and keep other nodes as it is, then it must become 2 rows. in such situation, backend will treat this as 2 different items rather than 1 item.

<br/>

feel free to let me know if I did not answer your question 🙏

steven-chiu-bigstack: Thanks for the clarification! @jdjgya 🙇‍♂️ 

I noticed Raven mentioned "Maybe in some cases...", so it seems like the behavior might not have occurred. Perhaps it was just a question of whether that behavior could happen under certain conditions.

the row would split into multiple rows if only some hosts get updated. But that would only happen on the Edit Tunings page where we can change the values for specific hosts, right? For enable/disable/reset, the backend ensures that once the API responds with a success status code, the single row with the same tuning name and host names will definitely present in the next listTunings API call (assuming we call it right after the mutation API).

jdjgya: Hi @steven-chiu-bigstack , Yes - that's right 👍

steven-chiu-bigstack: I just checked in with Raven, and he suggested that we should not use `pageNum` and `pageSize` when calling the API to ensure the target tuning is always included in the response when tuning count becomes larger. Thank you so much for catching this problem @raven-pan! 🙏 🙇‍♂️ 

By the way @jdjgya, I noticed the position (page) of a tuning will change after enable/disable. I haven't tried reset, but I think it's likely to happen to reset as well:

https://github.com/user-attachments/assets/5fc9e5d0-d5db-4de2-be93-ea28ae60d69b

It would be great if you could take a look at it when you have time, thank you! 🙇‍♂️

jdjgya: Hi @steven-chiu-bigstack , 

Thanks for the friendly reminder! yup -  this is a known symptom,  I'm still thinking the way to mitigate it. QQ

PR #473, opened 2025-05-22
Fix/99 not able to download support files

### LINKED ISSUE #99: [Bug]: Support file - Browser / UI is not able to download the support file by the button
## Describe the bug

Browser / UI has a chance to be not able to download the support file by the button somehow (like 60~70% would be blocked/failed from UI), 

but it works 100% if we just put the download url in the browser directly 🤔 

also, when the such symptom occurred, we can't see any api calls from UI / Browser by checking in the inspection mode.

<br/>

## Version and Env

Cube COS 3.0.0 RC1 (10.32.45.10)

<br/>

## To Reproduce

1). Go to support file page

2). Create a support file item or just pick an existing one, then download the support file.

3). Browser shows the `Check connect error` and failed to download

Image

<br/>

## Video

Some times it works, but feel high chance to get failure

https://github.com/user-attachments/assets/e59e0da0-0a71-4334-b843-a62efde1a96f

raven-pan:
#### What type of PR is this?

Fix

Fixes https://github.com/bigstack-oss/cubecos/issues/99

Use `window.open(url, '_blank')` instead of `<a href={url} target="_blank">` to ensure the browser opens a new tab to download support files.

Reference: https://stackoverflow.com/questions/29444051/link-with-target-blank-does-not-open-in-new-tab-in-chrome

#### Test Screenshots

Local Testing

https://github.com/user-attachments/assets/b06573a5-5570-467d-b852-526dfab8f0ff

This issue cannot be reproduced in the local dev environment, so I performed a simple test on https://10.32.45.10/maintenance/support-files

https://github.com/user-attachments/assets/9db85d38-e372-47f0-ab53-83db04d9116e

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/maintenance/supportFiles/_components/DownloadSupportFilesModal.tsx
+++ b/packages/cube-frontend-web-app/src/pages/maintenance/supportFiles/_components/DownloadSupportFilesModal.tsx
@@ -2,7 +2,6 @@ import { useState } from 'react'
 import { CosCheckbox, CosModal } from '@cube-frontend/ui-library'
 import { SupportFile } from '@cube-frontend/api'
 import { SupportFileRow } from '../MaintenanceSupportFilesPage'
-import { download } from '@cube-frontend/web-app/utils/download'
 import { formatSupportFilesTimestamp } from '@cube-frontend/web-app/utils/date'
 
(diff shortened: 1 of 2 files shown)


steven-chiu-bigstack (approved):
This browser behavior is so strange. 🤷‍♂️
Thanks for the effort, great fix! 👍🙏 🚀

PR #474, opened 2025-05-23
fix(web-app): license table support plan casing

steven-chiu-bigstack:
#### What type of PR is this?

Fix

Fix License Page support plan casing (should be upper case).

Close https://github.com/bigstack-oss/cubecos/issues/116#event-17782509570

image


raven-pan (approved):
LGTM

PR #472, opened 2025-05-22
469 fix: node resource usage percentage

steven-chiu-bigstack:
#### What type of PR is this?

fix

#### What this PR does / why we need it

- Adjust Node tables and Node Detail resource usage percentage source (use the percentage returned by the API instead of calculating it in the frontend).
- Enhancement:Show at least 1% in CosProgessBar when progress > 0.

   image

   - Before

      image

   - After

      image

#### Which issue(s) this PR fixes

Close bigstack-oss/cubecos#112

PR #475, opened 2025-05-23
Fix/469 chart percentage should use api response percentage

raven-pan:
#### What type of PR is this?

Fix

Adjust VM Allocation Chart Panel percentage source (use the percentage returned by the API instead of calculating it in the frontend).

Related: https://github.com/bigstack-oss/cubecos/issues/112

This PR reviews all related charts to ensure they use the percentage provided by the API rather than performing calculations on the frontend.

### DIFF
--- a/packages/cube-frontend-ui-library/src/components/CosPercentagePieChart/CosPercentagePieChart.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosPercentagePieChart/CosPercentagePieChart.tsx
@@ -9,6 +9,10 @@ export type CosPercentagePieChartProps = {
   total: number
   used: number
   /**
+   * If not provided, the percentage will be calculated based on used and total.
+   */
+  percentage?: number
+  /**
    * @default 100
    */
   thresholdPercentage?: number
@@ -43,6 +47,7 @@ export const CosPercentagePieChart = (props: CosPercentagePieChartProps) => {
     unit,
     total,
     used,
+    percentage: percentageProp,
     color: colorProp,
     thresholdPercentage = 100,
     percentageFormatter = defaultPercentageFormatter,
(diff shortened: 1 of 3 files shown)


packages/cube-frontend-ui-library/src/components/CosPercentagePieChart/CosPercentagePieChart.tsx
```diff
@@ -51,7 +56,12 @@ export const CosPercentagePieChart = (props: CosPercentagePieChartProps) => {
   } = props
 
   // Default to 0 to avoid NaN when total is 0.
-  const percentage = Math.floor((used / total) * 100) || 0
+  const getPercentage = () => {
+    const percentage = percentageProp ?? ((used / total) * 100 || 0)
+    return Math.floor(percentage)
```
steven-chiu-bigstack: In [yesterday's PR](https://github.com/bigstack-oss/cube-cos-ui/pull/472) for the node resource usage progress bars, I adjusted the code so that if `usedPercent` is between 0 and 1 (exclusive), the UI shows 1% instead of 0%. That's because the percentage in `CosProgressBar` is always an integer by design, and it feels a bit weird to see 0% when you know it's not zero.

image

image

Do you think we should do the same here? 🤔

steven-chiu-bigstack (approved):

packages/cube-frontend-ui-library/src/components/CosPercentagePieChart/CosPercentagePieChart.tsx
raven-pan: Sure! that wil be great!

raven-pan: I added a story to test this scenario.

Screenshot 2025-05-23 at 2 39 15 PM

PR #489, opened 2025-05-30
chore(release): sync branch release_3.0.0 with branch develop

Eandalf-Bigstack:
#### What type of PR is this?

- chore

- Sync branch release_3.0.0 with branch develop

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,6 +20,7 @@ updates:
       day: monday
       time: "09:00"
       timezone: "Asia/Taipei"
+    open-pull-requests-limit: 0
     groups:
       minor-and-patch:
         applies-to: version-updates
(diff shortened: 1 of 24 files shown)


steven-chiu-bigstack:
Hi @Eandalf-Bigstack, thanks for creating this PR! However, I'm a bit hesitant about syncing `release_3.0.0` with `develop` right now. The initial plan for the 3.0 release was to only include changes from M1. We've introduced some M2 refactors to `develop` recently, which likely won't impact functionality, but it's probably better to keep `release_3.0.0` away from M2 changes until M1 is officially ended.
Have the RC builds officially transitioned to the stable build yet? If so, I think we can proceed and merge this. Otherwise, does it make sense to hold this PR until M1 is wrapped up? What do you think? 🤔

Eandalf-Bigstack:
I thought we were missing some commits for branch release_3.0.0. If these commits are for M2, then we could close this PR. Since we are having GA in a couple few days, I was checking all stuff is merged into RC2 first, but if these are not related to M1, then this PR is not relevant anymore.

PR #508, opened 2025-06-06
506 settings page email recipients bug

### LINKED ISSUE #506: [Bug]: isUpdating Spinner Not Automatically Shown After Creating Email Recipients
## Describe the bug

When creating Email Recipients, it does not automatically enter the isUpdating loading state — a refresh is required for it to display.

## Version

COS 3.0.0 - RC3

## To Reproduce

1. Go to Setting page
2. Create Email Recipients (Slack and the Email Sender too)
3. The UI did not enter the isUpdating loading spinner state

## Expected behavior

After creation, the isUpdating loading spinner should be displayed automatically.

## Screenshots

Image

steven-chiu-bigstack:
#### What type of PR is this?

Fix

On Settings Page:

- Fix created row is not in a loading status by default bug.
- Fix deleted row is not in a loading status by default bug.
- Fix deleted row won't be removed from UI during polling bug.

Close #506

- Create email recipient

   https://github.com/user-attachments/assets/160d3f37-28fc-4f83-8698-48ea5d21c385

- Delete email recipient

   https://github.com/user-attachments/assets/8169c50e-e9af-43ed-b506-d585c1f0978c

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/settings/_components/email/recipients/actions/createEmailRecipient.ts
+++ b/packages/cube-frontend-web-app/src/pages/settings/_components/email/recipients/actions/createEmailRecipient.ts
@@ -28,12 +28,11 @@ export const createEmailRecipient = async (
     patchRow(row.id, {
       id: getRowId(),
       originalState: newEmailRecipient,
-      status: {
-        current: SettingStatusCurrentEnum.Ok,
-        isUpdating: false,
-      },
       isNew: false,
       isEditing: false,
+      // Don't set `status.isUpdating` to false after creation because
+      // email recipients are in an updating state by default upon creation.
+      // Status will be synced by polling later.
     })
     onSuccess?.()
   } catch (error) {
--- a/packages/cube-frontend-web-app/src/pages/settings/_components/email/recipients/useEmailRecipientRows.ts
+++ b/packages/cube-frontend-web-app/src/pages/settings/_components/email/recipients/useEmailRecipientRows.ts
@@ -1,4 +1,7 @@
-import { EmailRecipientResponse } from '@cube-frontend/api'
+import {
+  EmailRecipientResponse,
+  SettingStatusCurrentEnum,
+} from '@cube-frontend/api'
 import { DeepPartial } from '@cube-frontend/utils'
 import { DataCenterContext } from '@cube-frontend/web-app/context/DataCenterContext'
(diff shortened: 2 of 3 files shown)


raven-pan (commented):
I found a issue:

https://github.com/user-attachments/assets/b5c1b937-ef0d-4183-882c-7ea6b44115c3

The recipient `123@bigstack.co` disappears after polling.

steven-chiu-bigstack:
Good catch, thanks @raven-pan! Let me try to fix it. 🙏

steven-chiu-bigstack:
Update: It seems that there's an issue on the backend side. If multiple email recipients are added in a short time, only some of the records will be returned by the Get Settings API.

In the video below, `ajergsen@gmail.com`, `bjergsen@gmail.com`, and `cjergsen@gmail.com` are added in sequence, but it looks like the response only shows one email recipient being created at a time.

https://github.com/user-attachments/assets/776d452a-239a-4927-a023-91afe9e1f53f

However, if we wait long enough (e.g. 3 minutes), all 3 email recipients will eventually be included in the response:

image

We need @jdjgya's help with this one. 🙏 🙇‍♂️ 

---

Similar situation also happens to deletion. In the API response, only 1 email recipient will be in the updating status at a time, even if we know there should be more.

https://github.com/user-attachments/assets/6d48a5eb-eeb8-48c3-95ae-d1d0febf8b09

jdjgya:
Hi @steven-chiu-bigstack ,

<br/>

Thanks for the reminder. I suppose there're still some transition defects that BE overlooked during testing,

<br/>

I suppose this FE PR should be fine to merge first,

but for the backend transition part, the impacted scope of fix is not little, consider the final result will be fine to be applied in the cos (just the status transition defect), so should be handled in the M2 rather than this moment according the strategy we made recently.

raven-pan (approved):
LGTM.

steven-chiu-bigstack:
Really appreciate your help @jdjgya @raven-pan! 🙏 🎸 🙇‍♂️ 

I did a workaround to prevent table rows from vanishing abruptly during polling (found by @raven-pan 🚀). In the video below, the API response only returns a part of the email recipients, but the ones we just created in the UI are still visible:

https://github.com/user-attachments/assets/d86e3411-d34e-4867-9942-cae4899c6eca

Their statuses will still be synced as long as users don't reload the page:

{1EC7604D-3AF1-4565-9BA7-332CAF503403}

I think this is the best we could do for now.

PR #514, opened 2025-06-13
docs(readme): add repo structure, links to docs, simple set up, and code owners

Eandalf-Bigstack:
#### What type of PR is this?

- docs

- Add the documentation regarding the repository structure, links to documents in projects, and the simple set up
- Add code owners section to `.github` to automatically append the reviewer when creating a PR

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @steven-chiu-bigstack @raven-pan @lizzy-liang-bigstack
+* @bigstack-oss/front-end
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @steven-chiu-bigstack @raven-pan @lizzy-liang-bigstack
--- a/README.md
+++ b/README.md
@@ -1,5 +1,65 @@
 # CubeCOS UI
 
+## Repository Structure
+
+The repository follows the monorepo approach. Projects share the common utilities and the UI library.
+
+- [cube-frontend-utils](./packages/cube-frontend-utils/): The helper functions.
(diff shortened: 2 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @steven-chiu-bigstack @raven-pan @lizzy-liang-bigstack
```
steven-chiu-bigstack: Thanks @Eandalf-Bigstack! 🙏 

From [the docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) it looks like a team can also be a code owner:

> The people you choose as code owners must have write permissions for the repository. When the code owner is a team, that team must be visible and it must have write permissions, even if all the individual members of the team already have write permissions directly, through organization membership, or through another team membership.

This makes things slightly easier as we wouldn't need to update this file every time someone joins the frontend team. Something like @bigstack-oss/front-end should be working, but I'm not sure if the frontend team has the necessary write permissions.

I also noticed Gisele is part of the dev/frontend team. While the risk of non-developers accidentally approving unreviewed PRs is very, very low, it is still possible if the entire team is assigned as the code owner.

For now, I think explicitly listing everyone in this file is fine. We'll just need to remember to update it as needed in the future. 👌

steven-chiu-bigstack (approved):
LGTM, thanks @Eandalf-Bigstack! 🚀🚀🚀

.github/CODEOWNERS
Eandalf-Bigstack: Great point! I would update the setting to be set to @bigstack-oss/front-end. Currently, @bigstack-oss/front-end is a sub-team of team @bigstack-oss/dev, and team @bigstack-oss/dev already has the write permission to all repositories. I would remove @giselech from @bigstack-oss/front-end to keep the code owner list correct.

PR #517, opened 2025-06-16
fix(ui-library): 505 incorrect arrows_expand_02 icon display size

### LINKED ISSUE #505: UI Library - the arrows_expand_02 icon incorrect size, should aligned with the design guideline
Current Implementation (Incorrect):

Image

COS Design Guideline:

Image

[Figma](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=17-568&m=dev)

Notes:
1. Currently, this icon is not used in the COS UI.

raven-pan:
#### What type of PR is this?

Fix

Root Cause:

The `arrows_expand_02` icon was exported from Figma using the `SVG Exporter` plugin. we accidentally included the outer frame, which caused the icon to display incorrect size.

Solution:

Re-export the `arrows_expand_02` icon without the outer frame.

Before:
image

After:
image

Fixes #505


steven-chiu-bigstack (approved):
↗️ 
↙️

PR #522, opened 2025-06-16
perf: 442 apply useMemo in react context provider to prevent rerender

### LINKED ISSUE #442: [Performance] Use useMemo in React Context Provider
Image

raven-pan:
#### What type of PR is this?

Performance Impovement

perf: 442 apply useMemo in react context provider to prevent re-render.

Please see [React official docs example](https://react.dev/reference/react/useContext#optimizing-re-renders-when-passing-objects-and-functions).

Fixes #442


steven-chiu-bigstack (approved):
LGTM!

PR #524, opened 2025-06-16
feat(ui-library): create CosMessage component and its story

### LINKED ISSUE #19: UI Library - Message
[Mockup](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=486-8550&t=LfGu7yUKAaYr23yf-0)

lizzy-liang-bigstack:
feat: update story

#### What type of PR is this?

- Feature

- Create `CosMessage`

Screenshot 2025-06-16 at 4 26 06 PM

Fixes #19 

> Note

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-ui-library/src/components/CosMessage/CosMessage.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosMessage/CosMessage.tsx
@@ -1,0 +1,56 @@
+import { CosIconText } from '../CosIconText/CosIconText'
+
+type CosMessageProps = {
+  label?: string
+  time?: string
+  tag?: string
+  title: string
+  children: string
+}
+
+export const CosMessage = (props: CosMessageProps) => {
+  const { children, label, time, tag, title } = props
+
+  const renderLabel = () => {
+    return (
+      <div className="secondary-body2 font-semibold text-functional-text">
+        {label}
+      </div>
+    )
+  }
+
+  const renderTime = () => {
+    if (!time) return null
+    return (
+      <div className="secondary-body5 text-functional-text-light">
+        Time: {time}
+      </div>
+    )
+  }
+
+  const renderHeader = () => {
+    if (!label && !time) return null
+    return (
+      <div className="flex items-center justify-between">
+        {renderLabel()}
+        {renderTime()}
+      </div>
+    )
+  }
+
+  const renderTag = () => {
+    if (!tag) return null
+    return <CosIconText type="primary">{tag}</CosIconText>
+  }
+
+  return (
+    <div className="flex flex-col gap-y-1">
+      {renderHeader()}
+      <div className="flex items-center gap-3 rounded-[5px] bg-secondary-50 px-4 py-[14px]">
+        {renderTag()}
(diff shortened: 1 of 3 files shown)


packages/cube-frontend-ui-library/src/components/CosMessage/CosMessage.tsx
```diff
@@ -0,0 +1,58 @@
...
+        {renderLabel()}
+        {renderTime()}
+      </div>
+    )
+  }
+
+  const renderTag = () => {
+    if (!tag) return null
+    return (
+      <div className="primary-body5 rounded bg-primary px-1 py-0.5 font-semibold text-grey-0">
```
steven-chiu-bigstack: Nitpick: We actually have a component for the tag:

```tsx
<CosIconText type="primary">{tag}</CosIconText>
```

The name is indeed not straightforward. 😕 It is named after the [mockup](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=1-140&m=dev) (Icons -> Text):

{856D8A24-9470-4689-BFB2-AB80AC04082E}

packages/cube-frontend-ui-library/src/components/CosMessage/CosMessage.tsx
```diff
@@ -0,0 +1,58 @@
...
+        {tag}
+      </div>
+    )
+  }
+
+  return (
+    <>
```
steven-chiu-bigstack: It is suggested to replace the root fragment with a `div`. This ensures consistent spacing between the header and message content, even when a Y-gap class (such as `flex flex-col gap-y-4`) is defined in the parent.

This also means we can remove `mb-1` from the header and move the Y-gap class to the root `div`.

packages/cube-frontend-ui-library/src/components/CosMessage/CosMessage.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <div className="secondary-body2 font-semibold text-functional-text">
+        {label}
+      </div>
+    )
+  }
+
+  const renderTime = () => {
+    if (!time) return null
+    return (
+      <div className="secondary-body5 justify-end text-functional-text-light">
```
steven-chiu-bigstack: It looks like we can remove `justify-end` from here because `justify-between` is applied to the header container.

steven-chiu-bigstack (approved):
LGTM!

packages/cube-frontend-ui-library/src/components/CosMessage/CosMessage.tsx
lizzy-liang-bigstack: You're right! I'll use `CosIconText` instead!

packages/cube-frontend-ui-library/src/components/CosMessage/CosMessage.tsx
lizzy-liang-bigstack: Ah, good catch! Thanks for pointing that out! 🙏

PR #535, opened 2025-06-20
perf: 530 debounce the tunings page search requests

### LINKED ISSUE #530: [Performance] Debounce the Tunings Page Search Requests
### Problem

The COS UI sends a request every time the user types a character.

https://github.com/user-attachments/assets/0cf922d6-1cfb-4ec1-a865-cd12743da4b3

### Solution

We can use debounce to reduce network traffic.

raven-pan:
#### What type of PR is this?

Performance

perf: debounce the tunings page search requests.

https://github.com/user-attachments/assets/17a0a880-b89e-4acd-91b2-3251dad64575

Fixes #530

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/events/tunings/EventsTuningsPage.tsx
+++ b/packages/cube-frontend-web-app/src/pages/events/tunings/EventsTuningsPage.tsx
@@ -28,6 +28,7 @@ const TuningTable = GetCosBasicTable<TuningRow>()
 export const EventsTuningsPage = () => {
   const {
     query,
+    keywordDebouncedQuery,
     onKeywordChange,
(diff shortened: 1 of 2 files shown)


steven-chiu-bigstack (approved):
LGTM!

After all higher priority tasks are done, we can take a deeper look at the filter and periodic-fetch combination to see if there's an elegant solution for applying the debounce mechanism to all filter properties, not just keyword.

raven-pan:
> LGTM!
> 
> After all higher priority tasks are done, we can take a deeper look at the filter and periodic-fetch combination to see if there's an elegant solution for applying the debounce mechanism to all filter properties, not than just keyword.

Sure!

PR #538, opened 2025-06-24
516 node ipmi control UI

### LINKED ISSUE #516: [Page]: Node IPMI Control UI

Implement node IPMI control UI.

Implement node IPMI control UI.

**Additional context**

- [Mockup](https://www.figma.com/design/kSUzxXJ3OZ5RXSZCRjZ9o1/COS-3.0.1-UI-M2-mockup?node-id=12006-31773&t=L7oA0We7fQTfgmIg-0)
- The change of [action menu options](https://www.figma.com/design/kSUzxXJ3OZ5RXSZCRjZ9o1/COS-3.0.1-UI-M2-mockup?node-id=12006-31386&t=L7oA0We7fQTfgmIg-0) will be handled in bigstack-oss/cube-cos-ui#438

steven-chiu-bigstack:
#### What type of PR is this?

Feat

Implement Node IMPI control UI.

Fixes #516

- Connecting (enabling IPMI control)

   https://github.com/user-attachments/assets/819479db-b6a9-4914-b76c-66149d6829cd

- Disconnecting

   https://github.com/user-attachments/assets/1177bbd2-5c99-47d3-8b7a-467949fae5f7

- Validation failed

https://github.com/user-attachments/assets/f8eac003-10b9-452d-a23d-7368e1ea2f5b

- New action menu items once IPMI is turned on

{57FF9165-EE9F-4149-8AD0-09A095D294F7}


packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeDetailsHeader.tsx
```diff
@@ -62,24 +103,34 @@ export const NodeDetailsHeader = (props: NodeDetailsHeaderProps) => {
...
+          Component: Link,
+          props: {
+            to: CosRoutesEnum.NODES_PAGE,
+          },
+        }}
+      >
+        {node?.hostname ?? ''}
+      </CosBackButton>
+      {node?.ipmi.isSupported && (
+        <Link to={CosRoutesEnum.NODE_IPMI_CONTROL_PAGE(node?.hostname)}>
```
Comment (raven-pan) [inline packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeDetailsHeader.tsx]: nitpick: We already checked that `node` is not null in the previous line, so we can safely remove the `?` optional operator.

packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeSummary.tsx
```diff
@@ -117,13 +121,21 @@ export const NodeSummary = (props: NodeSummaryProps) => {
   }
 
   const getLicenseExpiration = (): string => {
```
Comment (raven-pan) [inline packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeSummary.tsx]: Maybe we could reuse `toLicenseExpirationDate` fn in `utils/date.ts`?

packages/cube-frontend-web-app/src/pages/node/[name]/_components/useIPMIOperations.ts
```diff
@@ -0,0 +1,73 @@
...
+  const [isIPMIOperating, setIsIPMIOperating] = useState(false)
+
+  const operate = async (
+    operation: OperateNodeIpmiOperationEnum,
+  ): Promise<void> => {
+    setIsIPMIOperating(true)
+    try {
+      await nodesApi.operateNodeIpmi({
+        dataCenter: dataCenter!.name,
+        nodeName: node?.hostname ?? '',
```
Comment (raven-pan) [inline packages/cube-frontend-web-app/src/pages/node/[name]/_components/useIPMIOperations.ts]: Discussion:
Maybe we can throw an error or early return if `nodeName` is `null` or an empty string before the API call.

packages/cube-frontend-web-app/src/pages/node/[name]/ipmi-control/IPMICancelButton.tsx
```diff
@@ -0,0 +1,22 @@
+import { CosButton } from '@cube-frontend/ui-library'
+import { CosRoutesEnum } from '@cube-frontend/web-app/enum/routes'
+import { useNavigate } from 'react-router'
+
+type IPMICancelButtonProps = {
+  nodeName: string
+}
+
+export const IPMICancelButton = (props: IPMICancelButtonProps) => {
```
Comment (raven-pan) [inline packages/cube-frontend-web-app/src/pages/node/[name]/ipmi-control/IPMICancelButton.tsx]: This component can be safely removed.

raven-pan (approved):

raven-pan (commented):
The Status Component may need more width to display correctly

Screenshot 2025-06-30 at 6 59 39 PM

raven-pan (commented):
Discussion:
Should we disable the `Create Support Files` operation for power-off nodes?

Screenshot 2025-06-30 at 6 54 22 PM

Screenshot 2025-06-30 at 6 54 02 PM

In the current implementation, the COS API will accept the operation, but no support file will be created.

https://github.com/user-attachments/assets/aee3dedc-6413-4368-9e8e-e18174e1b3c9

raven-pan (commented):
Some times the Node COS API responds with a  `""` status, which is not expected.

Screenshot 2025-06-30 at 7 03 59 PM

raven-pan (commented):
On the License Page, the Copy Serial Numbers modal shows `unlicensed` for powered-off Cube453 nodes. Maybe we can discuss and find a way to prevent user confusion.

Screenshot 2025-06-30 at 6 55 19 PM

raven-pan (commented):
image

On the IPMI Control page, the validation always passes regardless of which port the user specified. Is that expected?

We can discuss this with the COS API developers.

packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeSummary.tsx
Comment (steven-chiu-bigstack) [reply packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeSummary.tsx]: I did think about reusing `toLicenseExpirationDate`, but `toLicenseExpirationDate` formats dates to `YYYY/MM/DD`, while [the one](https://www.figma.com/design/kRJ5MUAkMF0AYgXCeChtZn/COS-management-UI-Redesign-v.3?node-id=3281-15059&m=dev) in this page has to be `YYYY/MM/DD HH:mm`.

{5E14DE0B-1DF2-40D8-BECA-9C6B4A7C0E02}

Do you think it's a good idea to add an argument to `toLicenseExpirationDate` that determines whether to include `HH:mm` in the result? (sounds good to me!)

packages/cube-frontend-web-app/src/pages/node/[name]/_components/useIPMIOperations.ts
Comment (steven-chiu-bigstack) [reply packages/cube-frontend-web-app/src/pages/node/[name]/_components/useIPMIOperations.ts]: Absolutely. Now that you mention it, the non-null assertion operator `!` also looks reasonable to me, just like how we did it with `dataCenter!.name`. Do you think it make sense to use `node!.hostname` here? (it is to me)

packages/cube-frontend-web-app/src/pages/node/[name]/ipmi-control/IPMICancelButton.tsx
Comment (steven-chiu-bigstack) [reply packages/cube-frontend-web-app/src/pages/node/[name]/ipmi-control/IPMICancelButton.tsx]: Sorry, my bad. 🙏

steven-chiu-bigstack:
> The Status Component may need more width to display correctly

You're right! I'll add `whitespace-nowrap` to `CosStatus` so that the text doesn't wrap:

{80274F48-B62B-4A0A-8665-B4BD56B91C2F}

> Discussion:
> Should we disable the Create Support Files operation for power-off nodes?

Good catch! Create support files operation should be disabled for any node that's not in `up` status. I've also updated the Node List page (by hiding the checkboxes of non-up nodes) to align with this change.

However, the built-in disabled style from the batch action table looks a bit weird when non-text content are involved, so I adjusted  the styles to make it look slightly more natural:

image

There'll be some edge cases where a node might be `up` when we check it, but transition to `power off` right after. I haven't handled this on the UI side because the API should always perform this check no matter what. When it happens, we can simply show a toast notification based on the API's response (and no support files will be created.)

> Some times the Node COS API responds with a "" status, which is not expected.

I've ran into this problem a few times, it's rare though. We can inform backend developers about this.

> On the License Page, the Copy Serial Numbers modal shows unlicensed for powered-off Cube453 nodes. Maybe we can discuss and find a way to prevent user confusion.

It looks like license status might be unavailable for down nodes. We can confirm this with the backend developers and see what to do next. 🙏 

> On the IPMI Control page, the validation always passes regardless of which port the user specified. Is that expected?
> We can discuss this with the COS API developers.

My guess is the validation should fail when the port is incorrect. Let's check in with the backend developers.

packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeSummary.tsx
Comment (raven-pan) [reply packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeSummary.tsx]: Yes, Sure!  the `Unlicense` and empty string display logic seems easy to lead new developers to make mistakes.

packages/cube-frontend-web-app/src/pages/node/[name]/_components/useIPMIOperations.ts
Comment (raven-pan) [reply packages/cube-frontend-web-app/src/pages/node/[name]/_components/useIPMIOperations.ts]: Sure!

steven-chiu-bigstack:
Going to merge this PR now, really appreciate your detailed review @raven-pan! 🚀

PR #543, opened 2025-06-24
chore(web-app): deploy local dist to dev remote scripts

raven-pan:
#### What type of PR is this?

Chore

Please see `packages/cube-frontend-web-app/README.md`.

### DIFF
--- a/packages/cube-frontend-web-app/package.json
+++ b/packages/cube-frontend-web-app/package.json
@@ -9,7 +9,9 @@
     "dev": "vite",
     "build": "tsc -b && vite build",
     "preview": "vite preview",
-    "test": "vitest run"
+    "test": "vitest run",
+    "deploy-dev": "./scripts/deployToDevRemote.sh",
+    "predeploy-dev": "pnpm build"
   },
   "dependencies": {
     "@cube-frontend/api": "workspace:*",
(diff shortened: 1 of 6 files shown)


packages/cube-frontend-web-app/package.json
```diff
@@ -9,7 +9,9 @@
     "dev": "vite",
     "build": "tsc -b && vite build",
     "preview": "vite preview",
-    "test": "vitest run"
+    "test": "vitest run",
+    "deploy-dev": "./scripts/deployToDevRemote.sh",
+    "predeploy-dev": "pnpm build"
```
steven-chiu-bigstack, line 14: It seems that the `predeploy-dev` command is not being used anywhere. Does it make sense to remove it and replace it with the `build` command?

steven-chiu-bigstack (approved):
Looks good! Thanks for creating this convenient script, so useful! 🙏 👍 💯

packages/cube-frontend-web-app/package.json
raven-pan: In package.json, the `predeploy-dev` script will always run before the `deploy-dev` script.
So if a user runs `pnpm deploy-dev`, it will first execute `pnpm build`, followed by `./scripts/deployToDevRemote.sh`

reference: https://docs.npmjs.com/cli/v8/using-npm/scripts#pre--post-scripts

PR #546, opened 2025-06-25
chore: specify node semver for develop env and ci

raven-pan:
#### What type of PR is this?

Chore

1. Specify the Node.js semver in Github workflows for better stability.
2. Add a `.nvmrc` file to allow developers to switch to the Node.js version used by `cube-cos-ui`


Eandalf-Bigstack:
Currently, the NodeJS version used to build cube-cos-ui and cube-cos-login (Keycloak) inside CubeCOS build jail is `22.14.0`. We could use this to make the version more aligned. Or, are we intended to use a newer version `22.17.0`?

steven-chiu-bigstack (approved):
As someone who know nothing about COS and its related code, I guess upgrading Node from 22.14.0 to 22.17.0 should be okay 🤔?

I like the idea of keeping Node versions aligned across all environments, as it makes the dev environment much closer to the other ones, and that usually leads to a more stable output and fewer headaches from different Node versions.

steven-chiu-bigstack:
By the way, in case we need it, here are the changelogs from Node 22.15.0 to 22.17.0:

- https://nodejs.org/en/blog/release/v22.15.0
- https://nodejs.org/en/blog/release/v22.16.0
- https://nodejs.org/ko/blog/release/v22.17.0

Eandalf-Bigstack:
Yes, I believe that NodeJS v22.17.0 should not break code working with NodeJS v22.14.0. NodeJS has been quite stable within a major version since NodeJS v12. I would update the version of NodeJS being used in CubeCOS build jail from v22.14.0 to v22.17.0.

raven-pan:
Discussion: how about this solution? it achieves single state of true and helps prevent the overhead of managing Node.js versions across different repos.

This may require help from CubeCOS developer 🙏.

Additional information:
May need to run `nvm install` first, and then `nvm use`. 

Screenshot 2025-06-26 at 12 55 12 PM

Eandalf-Bigstack:
Yes, we could delegate the version control to `.nvmrc` of `cube-cos-ui`. However, `nvm use` would not automatically pull down the required version. We would need to add a script to install the required version, but this would be handled by CubeCOS.

Eandalf-Bigstack:
BTW, the NodeJS version used by `skyline-console` is v12.22.12.

PR #547, opened 2025-06-25
feat(web-app): enable multi-select filters on Events page

### LINKED ISSUE #483: [UI library]: Table advanced filter format should be multi-select
## Describe the bug

Image

## Version

10.32.45.10

## To Reproduce

1. Go to 'Events'
2. Click on 'filter'

lizzy-liang-bigstack:
#### What type of PR is this?

- Feat

1. Define the `events` query schema using `zod`, and also update them to align with recent API changes
2. Update UI: replace single-select dropdowns with multi-select components for the following fields: `severity`, `category`, `host`, and `instance`

Screenshot 2025-06-25 at 5 21 56 PM

Fixes #483 

> Note

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/FilterDropdown.tsx
+++ b/packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/FilterDropdown.tsx
@@ -6,14 +6,22 @@ import { FilterOptions } from './useEventsQuery'
 type FilterDropdownProps<Key extends keyof FilterOptions> = {
   isLoading: boolean
   filterKey: Key
-  options: FilterOptions[Key][]
-  selectedValue: FilterOptions[Key] | undefined
-  onChange: (key: Key, value: FilterOptions[Key] | undefined) => void
+  options: FilterOptions[Key]
+  selectedValue: string[]
+  onFieldChange: <Key extends keyof FilterOptions>(
+    key: Key,
+    value: string,
+  ) => void
+  onFieldAllCheckChange: <Key extends keyof FilterOptions>(
+    key: Key,
+    value: FilterOptions[Key],
+  ) => void
+  onFieldClear: <Key extends keyof FilterOptions>(key: Key) => void
 }
 
 const filterOptionBySearchValue = <Key extends keyof FilterOptions>(
   searchValue: string,
-  options: FilterOptions[Key][],
+  options: FilterOptions[Key],
 ): string[] => {
   if (searchValue === '') return options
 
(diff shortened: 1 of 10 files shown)


packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/FilterDropdown.tsx
```diff
@@ -57,17 +82,17 @@ export const FilterDropdown = <Key extends keyof FilterOptions>(
...
       onSearchChange={handleSearchChange}
       onClearClick={handleClearClick}
-      disabled={false}
-      isLoading={isLoading}
+      onAllCheckChange={handleAllCheckChange}
     >
-      <CosDropdown.Trigger placeholder={upperFirst(filterKey)}>
-        {selectedItem?.[0] ?? undefined}
+      <CosDropdown.Trigger placeholder={filterLabel}>
+        {filterLabel ?? undefined}
```
steven-chiu-bigstack: Nitpick: `?? undefined` can be removed.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/useEventsQuery.ts
```diff
@@ -21,110 +19,126 @@ type UseEventsQuery = {
...
+      const isSelected = (currentValues as string[]).includes(value)
+
+      return {
+        ...prev,
+        [key]: isSelected
+          ? prev[key].filter((item) => item !== value)
+          : [...prev[key], value],
+        currentPage: 1,
+      }
+    })
```
steven-chiu-bigstack, line 80: I think it's okay to simplify this `setQuery(...)` call to the following:

```ts
setQuery((prev) => {
  const currentValues = prev[key] as string[]

  const nextValues = currentValues.includes(value)
    ? currentValues.filter((item) => item !== value)
    : [...currentValues, value]

  return {
    ...prev,
    [key]: nextValues,
    currentPage: 1,
  }
})
```

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/utils.ts
```diff
@@ -1,16 +1,77 @@
...
 
-const getValidType = (type: string): GetEventsTypeEnum => {
+const _eventsQuerySchema = paginationQuerySchema.extend({
+  [EventsParamKeyEnum.Type]: z
+    .enum(
+      Object.values(GetEventsTypeEnum) as [
+        GetEventsTypeEnum,
+        ...GetEventsTypeEnum[],
+      ],
+    )
```
steven-chiu-bigstack: It looks like we can replace `.enum` with [`.nativeEnum`](https://v3.zod.dev/?id=native-enums) (for both `[EventsParamKeyEnum.Type]` and `[EventsParamKeyEnum.Severity]`) for less code:

```ts
[EventsParamKeyEnum.Type]: z
    .nativeEnum(GetEventsTypeEnum)
    .default(GetEventsTypeEnum.System)
```

The resulting type will be slightly different, but the value is still correct type-wise.

- Before (with `.enum`)

  {D74C2988-5EFC-4B7A-90B1-BE20F99AC8F2}

- After (with `.nativeEnum`)

  {3CA202AE-D5E8-4E7A-9E15-A47DABAFF59D}

steven-chiu-bigstack (approved):
Hi @lizzy-liang-bigstack 

The type stuff is getting a little bit complicated, but the core functionality is working perfectly. We can revisit and refactor this part when we've got more bandwidth.

Thank you for taking care of this! 🙏 👍 🙇‍♂️

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTable/useEventsQuery.ts
lizzy-liang-bigstack: @steven-chiu-bigstack It's a great point! Since we are using `.nativeEnum` to define the severity type, it now correctly infers as `string[]`, so we no longer need to handle `GetEventsSeveritiesEnum` separately. Thank you for the clean solution :)

PR #567, opened 2025-07-14
566 refactor `CosDropdown`

### LINKED ISSUE #566: Refactor Dropdown component for better maintainability and type safety
### Problem Statement: 
1. Props are tangled, leading to complicated inner judgement.
2. While the current logic and styles work as expected, the internal structure lacks clarity, which increases maintenance cost.

### Proposed Scope:
1. Refactor the Dropdown component for clearer structure and better separation of concerns.
2. Maintain existing functionality and styling without any changes of UI behaviour.

lizzy-liang-bigstack:
#### What type of PR is this?

- Refactor

1. Refactor the `CosDropdown` component for a clearer and more maintainable structure
2. Rewrites its prop types to be more expressive and accurate
3. No visual redesign or behavioural change

Fixes #566

1. Prop type overhaul
    - Added `size` (`CosDropdownSize`, `md` | `sm`): use `sm` for `in-table` dropdown. This clarifies that the in-table version is just a size variant and not necessarily tied to a table component. 
    - Reassign `type` to `CosDropdownType` (`radio` | `checkbox`): for specifying selection type.
    - Reassign `variant` to `CosDropdownVariant` (`regular` | `withFilter`):  to distinguish between dropdowns with and without a filter bar.
    - Remove `searchValue` and `onSearchChange`: the filter input is now handled internally, as its logic is consistent across use cases.

2. Internal filter input handling
    - Filter input state is now managed within the component for simplicity and consistency.

3. Update stories and implementation
    - Updated the `CosDropdown` storybook and replaced all usage across `web-app`.

4. Removed duplicate `RoleDropdown`
    - Replaced it with the shared `RoleFilter` component for better reusability and consistency.

- Demo (take `License` for example):

https://github.com/user-attachments/assets/5475c7ee-86bd-4232-bcee-3fb21b71757e

#### Additional documentation

```docs

```


packages/cube-frontend-ui-library/src/components/CosDropdown/CosDropdownItem.tsx
```diff
@@ -1,67 +1,72 @@
...
     if (disabled) return
 
-    onClick()
-    if (type === 'regular' || type === 'search') {
-      toggleDropdownOpen()
-    }
-  }
+    onClickProp()
+    if (type === 'radio') toggleDropdownOpen()
+  }, [disabled, onClickProp, toggleDropdownOpen, type])
```
steven-chiu-bigstack: Nitpick: `onClickProp` is usually something that changes within every render, so `useCallback` may not be very useful here.

packages/cube-frontend-ui-library/src/components/CosDropdown/CosDropdownItem.tsx
```diff
@@ -1,67 +1,72 @@
...
 
-  const getCheckboxColor = (): CosCheckboxColor | undefined => {
-    if (type === 'checkbox') return 'primary'
-    if (type === 'search-checkbox') return 'secondary'
-    return undefined
+  if (variant === 'withFilter' && !shouldDisplay) {
+    return null
```
steven-chiu-bigstack, line 55: `shouldDsiplay` sounds like the item should've been hidden if it's `false` no matter what, but it's only made for `withFilter` dropdown only. Perhaps we can make use of a function?

```tsx
// Declaration
const shouldDisplay = (): boolean => {
  if (variant !== 'withFilter') {
    // Items in non-filter dropdowns should always be visible.
    return true
  }
  if (!searchValue) return true
  return label.toLowerCase().includes(searchValue.toLowerCase())
}

// And use it like this:
if (!shouldDisplay()) {
  return null
}
```

packages/cube-frontend-ui-library/src/components/CosDropdown/CosDropdownTrigger.tsx
```diff
@@ -17,42 +17,47 @@ export const CosDropdownTrigger = (props: CosDropdownTriggerProps) => {
...
 
   const renderSelectedItemCount = () => {
-    return type === 'search-checkbox' && isSelected ? (
-      <p>{`(${selectedItems.length})`}</p>
-    ) : null
+    if (type === 'radio' || !isSelected) return null
+    return (
+      <p
+        className={trigger.count({ variant, disabled })}
+      >{`(${selectedItems.length})`}</p>
```
steven-chiu-bigstack: Prettier sometimes run into this strange problem where it incorrectly formats HTML. 😅 
We can add a new line after the closing caret of `<p>` and format again to fix it:

{2D515FEB-1578-4200-BCF8-E740B089EA07}

packages/cube-frontend-ui-library/src/stories/components/CosDropdown/DropdownLayout.tsx
```diff
@@ -0,0 +1,44 @@
...
+  </>
+)
+
+export const DropdownRow = (props: {
+  size: CosDropdownSize
+  isLoading: boolean
+  selected: boolean
+  disabled: boolean
+  label?: string
+}) => {
```
steven-chiu-bigstack, line 35: Note (not a big deal, no need to make any changes): It looks like we're trying to proxy down the props. If `RadioDropdownProps` and `CheckboxDropdownProps` are both exported, we can use the `Pick` util type to pick those shared properties:

```tsx
type DropdownRowProps = Pick<
  RadioDropdownProps | CheckboxDropdownProps,
  'only' | 'shared' | 'props' | 'will-be-available' | 'here'
>
```

If they're not exported, the built-in `ComponentProps<T>` type would be helpful:

```tsx
import type { ComponentProps } from 'react'
import { RadioDropdown } from './RadioDropdown'

type RadioDropdownProps = ComponentProps<typeof RadioDropdown>
```

packages/cube-frontend-ui-library/src/stories/components/CosDropdown/CheckboxDropdown.tsx
```diff
@@ -1,183 +1,108 @@
...
+import { mockData, MockDataItem } from './mockData'
 
 type CheckboxDropdownProps = {
+  size: CosDropdownSize
   variant: CosDropdownVariant
-  isDisabled: boolean
-  hasDefaultValue: boolean
-  hasSearchbar: boolean
+  isLoading: boolean
+  selected: boolean
```
steven-chiu-bigstack, line 13: Nitpick: I think names like `(is)DefaultSelected` would be more appropriate, but this component is not that important (story-only), so never mind!

steven-chiu-bigstack (approved):
Hi @lizzy-liang-bigstack 

Excellent job, thanks for the great refactor! 🎉 🎊 🚀 👍 

Just a thought on the clear button of `CosDropdownTrigger`: I think it would be more intuitive to show it whenever it's a multi-select (checkbox) dropdown, not just when the variant is `withFilter`. But I understand that the current implementation matches the mockup exactly, so I guess it's fine to leave it as is. 🤷‍♂️

PR #570, opened 2025-07-17
569 Update Triggers List

### LINKED ISSUE #569: [Feature] Update Trigger List Page
[Mockup](https://www.figma.com/design/kSUzxXJ3OZ5RXSZCRjZ9o1/COS-3.1-M2-UI-mockup?node-id=12769-29279&m=dev)

lizzy-liang-bigstack:
⚠️ Note: Waiting for API and COS implementation to be completed.
⚠️ Currently use mostly mock data to build and test UI.

#### What type of PR is this?

- Feature

1. Update API schema
2. Update the trigger list UI accordingly:
    - Sort triggers by the `isBuiltIn` value (built-in triggers appear first).
    - Support three response types: `script`, `email`, and `slack`.
    - Add pagination support to the trigger list.
3. Removes outdated components related to trigger creation

Screenshot 2025-07-17 at 10 36 17 AM

Fixes #569 

> Note 💡 You can use `mockData` to preview a more complete trigger list.

```ts
// In `useTriggerRows.ts` file

 useEffect(() => {
    // const triggers = listTriggersResponse?.triggers ?? []
    const triggers = mockTriggers // ← Temporarily use mock data instead of API data

```

#### Additional documentation

```docs

```


steven-chiu-bigstack (approved):

packages/cube-frontend-web-app/src/pages/events/triggers/EventsTriggersPage.tsx
```diff
@@ -16,52 +23,74 @@ export const EventsTriggersPage = () => {
...
+        <TriggersTable rows={rows} isLoading={isLoading}>
+          <TriggersTable.Column
+            label="Triggers"
+            property="name"
+            emphasize={true}
+          >
+            {(name, row) => (
+              <div className="flex items-center gap-2">
+                {name}
+                {row.isUpdating && <CosLoadingSpinner variant="dot120" />}
```
raven-pan: Are there any concerns about using `row.isUpdating` to replace `row.status.isUpdating`?

raven-pan (approved):

packages/cube-frontend-web-app/src/pages/events/triggers/EventsTriggersPage.tsx
lizzy-liang-bigstack: Actually, you're right! We should use `row.status.isUpdating` instead of `row.isUpdating`. This would avoid the need for additional handling of `row.isUpdating`, and using `row.status.isUpdating` directly from the API would let the UI to better reflect the actual status. Thank you for pointing it out!

PR #571, opened 2025-07-17
554 Create Triggers Page

### LINKED ISSUE #554: [Feature] Create Triggers
Enable creating triggers from scratch.

As outlined in the [Mockup](https://www.figma.com/design/kSUzxXJ3OZ5RXSZCRjZ9o1/COS-3.1-M2-UI-mockup?node-id=12008-36110&m=dev).

－

**Additional context**
－

lizzy-liang-bigstack:
⚠️ This PR is based on #570 
⚠️ Note: Waiting for API and COS implementation to be completed.
⚠️ Currently use mostly mock data to build and test UI.

#### What type of PR is this?

- Feature

1. Select Events

https://github.com/user-attachments/assets/de2e4502-d780-4cf1-91f8-fd86dc44b040

2. Set Response

https://github.com/user-attachments/assets/a3f69cf9-45f7-4354-83fb-882d1f78a53e

3. Add Description

https://github.com/user-attachments/assets/15a9b79d-afbc-487b-aff7-0599b45dfa3e

Fixes #554

> Note

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-web-app/src/components/LogConsole.tsx
+++ b/packages/cube-frontend-web-app/src/components/LogConsole.tsx
@@ -1,0 +1,68 @@
+import { twMerge } from 'tailwind-merge'
+import { CosHyperlink, CosTooltip } from '@cube-frontend/ui-library'
+
+export type LogConsoleProps = {
+  title: {
+    label: string
+    href?: string
+  }
+  children: string
+  copyTooltip?: string
+}
+
+export const LogConsole = (props: LogConsoleProps) => {
+  const { title, children, copyTooltip } = props
+
+  const onCopyClick = () => {
+    if ('navigator' in window && children) {
+      window.navigator.clipboard.writeText(children)
+    }
+  }
+
+  const renderTitle = () => {
+    if (title.href) {
+      return (
+        <CosHyperlink variant="text-only" href={title.href} target="_blank">
+          {title.label}
+        </CosHyperlink>
+      )
+    }
+    return (
+      <span className="primary-body2 font-medium text-primary">
+        {title.label}
+      </span>
+    )
+  }
+
+  return (
+    <div className="flex flex-col">
+      <div
+        className={twMerge(
+          'flex items-center justify-between px-6 py-3',
+          'rounded-t-[5px] border border-functional-border-divider bg-scene-background',
+        )}
+      >
+        {renderTitle()}
+        <CosTooltip
+          hoverContent={{
+            message: copyTooltip ?? 'Click to copy details',
+          }}
+          clickContent={{
+            message: 'Copied!',
+          }}
+        >
+          <button
+            type="button"
+            className="primary-body2 cursor-pointer font-semibold text-functional-text"
+            onClick={onCopyClick}
+          >
+            Copy
+          </button>
+        </CosTooltip>
+      </div>
+      <div className="primary-body3 overflow-x-auto text-wrap rounded-b-[5px] bg-dark-700 px-6 py-4 text-functional-border-darker">
+        {children}
+      </div>
+    </div>
+  )
+}
--- a/packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/AddDescription/AddDescription.tsx
+++ b/packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/AddDescription/AddDescription.tsx
@@ -1,0 +1,84 @@
+import {
+  CosButton,
+  CosInput,
+  CosStroke,
+  CosTextArea,
+} from '@cube-frontend/ui-library'
+import { StepBoard } from '@cube-frontend/web-app/components/StepBoard'
+import { UpsertTriggersPayload } from '../../upsertTriggersUtils'
+import { TriggersPreviousButton } from '../TriggersPreviousButton'
+import { twMerge } from 'tailwind-merge'
+
+type AddDescriptionProps = {
+  isLoading: boolean
+  /**
+   * @default false
+   */
+  isEditMode?: boolean
+  nextButtonText: string
+  payload: UpsertTriggersPayload
+  errorMessage?: string | undefined
+  onNameChange: (e: React.ChangeEvent<HTMLInputElement>) => void
+  onDescriptionChange: (e: React.ChangeEvent<HTMLTextAreaElement>) => void
+  onPublishClick: (payload: UpsertTriggersPayload) => void
+}
+
+export const AddDescription = (props: AddDescriptionProps) => {
+  const {
+    isLoading,
+    isEditMode = false,
+    nextButtonText,
+    payload,
+    errorMessage,
+    onNameChange,
+    onDescriptionChange,
+    onPublishClick: onPublishClickProp,
+  } = props
+
+  const isValueValid = !!payload.name
+
+  const onPublishClick = () => {
+    onPublishClickProp(payload)
+  }
+
+  return (
+    <div className="flex flex-col gap-4">
+      <StepBoard>
+        <CosInput
+          isLoading={isLoading}
+          label="Trigger Name (used for identification)"
+          placeholder="Name"
+          value={payload?.name}
+          onChange={onNameChange}
+          className={twMerge(
+            'max-w-[512px]',
+            isEditMode && 'cursor-not-allowed',
+          )}
+          disabled={isEditMode}
+        />
+        <CosTextArea
+          isLoading={isLoading}
+          label="Description"
+          placeholder="Description"
+          maxLength={100000000}
+          value={payload?.description}
+          onChange={onDescriptionChange}
+        />
+      </StepBoard>
+      <CosStroke type="dot" />
+      {errorMessage && (
+        <div className="primary-body3 text-status-negative">{errorMessage}</div>
+      )}
+      <div className="flex items-center gap-x-4">
+        <TriggersPreviousButton />
+        <CosButton
+          className="self-start"
+          disabled={!isValueValid}
+          onClick={onPublishClick}
+        >
+          {nextButtonText}
+        </CosButton>
+      </div>
+    </div>
+  )
+}
--- a/packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AddAttributeModal.tsx
+++ b/packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AddAttributeModal.tsx
@@ -1,0 +1,287 @@
+import { ReactNode, useCallback, useEffect, useMemo, useState } from 'react'
+import xor from 'lodash/xor'
+import { CosDropdown, CosModal } from '@cube-frontend/ui-library'
+import {
+  GetPredefinedEventsSeveritiesEnum,
+  GetPredefinedEventsTypesEnum,
+} from '@cube-frontend/api'
+import {
+  attributeLabelMap,
+  TriggerAttributeKeys,
+  TriggerAttributes,
+  UpsertTriggersPayload,
+} from '../../upsertTriggersUtils'
+import { AttributeCheckboxGroup } from './AttributeCheckboxGroup'
+import { isEqual } from 'lodash'
+
+const checkIsAllChecked = (
+  selectedAttributes: string[],
+  options: string[],
+): boolean | null => {
+  if (selectedAttributes.length === options.length) return true
+  if (selectedAttributes.length === 0) return false
+  return null
+}
+
+const checkIsSelectionChanged = (
+  dropdownOptions: TriggerAttributeKeys[],
+  defaultSelection: TriggerAttributes,
+  selection: TriggerAttributes,
+): boolean => {
+  return dropdownOptions.some((key) => {
+    const defaultValue = defaultSelection[key]
+    const selectedValue = selection[key]
+    return !isEqual(defaultValue, selectedValue)
+  })
+}
+
+type AddAttributeModalProps = {
+  isModalOpen: boolean
+  payload: UpsertTriggersPayload
+  attributes: TriggerAttributes
+  activeType: TriggerAttributeKeys | undefined
+  dropdownOptions: TriggerAttributeKeys[]
+  onModelClose: () => void
+  onActiveTypeChange: (type: TriggerAttributeKeys) => void
+  onActionClick: (attributes: TriggerAttributes) => void
+}
+
+export const AddAttributeModal = (props: AddAttributeModalProps) => {
+  const {
+    isModalOpen,
+    payload,
+    attributes,
+    activeType,
+    dropdownOptions,
+    onModelClose: onModelCloseProp,
+    onActiveTypeChange,
+    onActionClick: onActionClickProp,
+  } = props
+
+  const { alertTypes, severities, categories, eventIds } = attributes
+
+  const {
+    alertTypes: defaultSelectedAlertTypes,
+    severities: defaultSelectedSeverities,
+    categories: defaultSelectedCategories,
+    eventIds: defaultSelectedEventIds,
+  } = payload
+
+  // Clone values to avoid referencing the original payload fields directly
+  // Prevents unintended shared mutations in state
+  const cloneAttributes = useCallback((): TriggerAttributes => {
+    return structuredClone({
+      alertTypes: defaultSelectedAlertTypes,
+      severities: defaultSelectedSeverities,
+      categories: defaultSelectedCategories,
+      eventIds: defaultSelectedEventIds,
+    })
+  }, [
+    defaultSelectedAlertTypes,
+    defaultSelectedSeverities,
+    defaultSelectedCategories,
+    defaultSelectedEventIds,
+  ])
+
+  const [selectedAttributes, setSelectedAttributes] =
+    useState<TriggerAttributes>(cloneAttributes)
+
+  useEffect(() => {
+    setSelectedAttributes(cloneAttributes())
+  }, [cloneAttributes])
+
+  const isSelectionChanged = useMemo(() => {
+    const defaultAttributes = cloneAttributes()
+    return checkIsSelectionChanged(
+      dropdownOptions,
+      defaultAttributes,
+      selectedAttributes,
+    )
+  }, [cloneAttributes, dropdownOptions, selectedAttributes])
+
+  const onSelectedAlertTypeChange = (
+    alertType: GetPredefinedEventsTypesEnum,
+  ) => {
+    setSelectedAttributes((prev) => ({
+      ...prev,
+      alertTypes: xor(prev.alertTypes, [alertType]),
+    }))
+  }
+
+  const onAllSelectedAlertTypesChange = (): void => {
+    setSelectedAttributes((prev) => {
+      const updatedAlertTypes = checkIsAllChecked(prev.alertTypes, alertTypes)
+        ? []
+        : alertTypes
+
+      return { ...prev, alertTypes: updatedAlertTypes }
+    })
+  }
+
+  const onSelectedSeverityChange = (
+    severity: GetPredefinedEventsSeveritiesEnum,
+  ) => {
+    setSelectedAttributes((prev) => ({
+      ...prev,
+      severities: xor(prev.severities, [severity]),
+    }))
+  }
+
+  const onAllSelectedSeveritiesChange = (): void => {
+    setSelectedAttributes((prev) => {
+      const updatedSeverities = checkIsAllChecked(prev.severities, severities)
+        ? []
+        : severities
+
+      return { ...prev, severities: updatedSeverities }
+    })
+  }
+
+  const onSelectedCategoryChange = (category: string) => {
+    setSelectedAttributes((prev) => ({
+      ...prev,
+      categories: xor(prev.categories, [category]),
+    }))
+  }
+
+  const onAllSelectedCategoriesChange = (): void => {
+    setSelectedAttributes((prev) => {
+      const updatedCategories = checkIsAllChecked(prev.categories, categories)
+        ? []
+        : categories
+
+      return { ...prev, categories: updatedCategories }
+    })
+  }
+
+  const onSelectedEventIdChange = (eventId: string) => {
+    setSelectedAttributes((prev) => ({
+      ...prev,
+      eventIds: xor(prev.eventIds, [eventId]),
+    }))
+  }
+
+  const onAllSelectedEventIdsChange = (): void => {
+    setSelectedAttributes((prev) => {
+      const updatedEventIds = checkIsAllChecked(prev.eventIds, eventIds)
+        ? []
+        : eventIds
+
+      return { ...prev, eventIds: updatedEventIds }
+    })
+  }
+
+  const onModelClose = () => {
+    setSelectedAttributes({
+      alertTypes: defaultSelectedAlertTypes,
+      severities: defaultSelectedSeverities,
+      categories: defaultSelectedCategories,
+      eventIds: defaultSelectedEventIds,
+    })
+    onModelCloseProp()
+  }
+
+  const onActionClick = () => {
+    onActionClickProp(selectedAttributes)
+    onModelCloseProp()
+  }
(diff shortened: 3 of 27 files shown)


packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/UpsertTriggersSteps.tsx
```diff
@@ -0,0 +1,38 @@
...
+  },
+]
+
+export const UpsertTriggersSteps = (props: UpsertTriggersStepsProps) => {
+  const { step } = props
+
+  return (
+    <CosStepProcess>
+      {stepItems.map((stepItem, index) => (
+        <CosStepProcessItem
```
raven-pan, line 29: Nitpick: Use `CosStepProcess.Item` instead of `CosStepProcessItem`.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/UpsertTriggersSteps.tsx
```diff
@@ -0,0 +1,38 @@
...
+          key={stepItem.step}
+          label={stepItem.label}
+          serialNumber={index + 1}
```
raven-pan, line 32: It's not in the scope of this feature, but I believe we can rename this prop to something like `stepNumber` or `step`, since `serialNumber` sounds more like a machine or disk serial number, which could cause misunderstandings.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AddAttributeModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+          variant="regular"
+          isLoading={false}
+          disabled={false}
+          selectedItems={[activeType]}
+        >
+          <CosDropdown.Trigger>
+            {activeType ? attributeLabelMap[activeType] : 'Choose an attribute'}
+          </CosDropdown.Trigger>
+          <CosDropdown.Menu>
+            {dropdownOptions.map((option: TriggerAttributeKeys) => (
```
raven-pan: We can remove the `: TriggerAttributeKeys`.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AttributePanel.tsx
```diff
@@ -0,0 +1,235 @@
...
+  >([])
+
+  const checkIsAllSelected = () => {
+    const unselectedAttributeOptions = filterUnselectedAttributeOptions(payload)
+    return unselectedAttributeOptions.length === 0
+  }
+
+  const onActionClick = (attributes: {
+    alertTypes: GetPredefinedEventsTypesEnum[]
+    severities: GetPredefinedEventsSeveritiesEnum[]
```
raven-pan: nitpick: we can replace it by `TriggerAttributes` type.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AddAttributeModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+  const renderContent = activeType ? renderContentFnMap[activeType] : () => null
+
+  return (
+    <CosModal
+      isOpen={isModalOpen}
+      title="Add Attribute"
+      actionText="Set Response"
+      onActionClick={onActionClick}
+      onCloseClick={onModelClose}
+      className="h-[490px]"
```
raven-pan, line 257: We can disable the action button if no attributes are selected.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AttributePanel.tsx
```diff
@@ -0,0 +1,235 @@
...
+      />
+    )
+  }
+
+  const renderSeverityStackCard = () => {
+    if (payload.severities.length === 0) return null
+
+    return (
+      <TriggersStackCard
+        title="Severity"
```
raven-pan: nitpick: We can get the stack card title by `attributeLabelMap`.

packages/cube-frontend-web-app/src/components/UpsertTriggers/useCreateTriggersPayload.ts
```diff
@@ -0,0 +1,193 @@
...
+    emails: [],
+    slacks: [],
+    script: undefined,
+    name: '',
+    description: '',
+  })
+
+  const onAlertTypeSelect = (
+    alertTypes: GetPredefinedEventsTypesEnum[],
+  ): void => {
```
raven-pan, line 48: Question: Why we define void here? are there any concerns?

packages/cube-frontend-web-app/src/components/UpsertTriggers/upsertTriggersUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+    isEmpty(eventIds)
+  )
+}
+
+export const isResponseValid = ({
+  emails,
+  slacks,
+  script,
+}: UpsertTriggersPayload): boolean => {
+  return !(isEmpty(emails) && isEmpty(slacks) && !script)
```
raven-pan: We can just `return isEmpty(emails) || isEmpty(slacks) || !!script`.

packages/cube-frontend-web-app/src/components/UpsertTriggers/useCreateTriggersPayload.ts
```diff
@@ -0,0 +1,193 @@
...
+  onScriptChange: (file: ScriptFile) => void
+  onScriptRemove: () => void
+  onNameChange: (e: React.ChangeEvent<HTMLInputElement>) => void
+  onDescriptionChange: (e: ChangeEvent<HTMLTextAreaElement>) => void
+  onEventsReset: () => void
+  onResponseReset: () => void
+}
+
+export const useCreateTriggersPayload = (): UseCreateTriggersPayload => {
+  const [payload, setPayload] = useState<UpsertTriggersPayload>({
```
steven-chiu-bigstack: Nitpick: We can use a initializer function to initialize the value of the `payload` state.

packages/cube-frontend-web-app/src/components/UpsertTriggers/useCreateTriggersPayload.ts
```diff
@@ -0,0 +1,193 @@
...
+    script: undefined,
+    name: '',
+    description: '',
+  })
+
+  const onAlertTypeSelect = (
+    alertTypes: GetPredefinedEventsTypesEnum[],
+  ): void => {
+    setPayload((prev) => {
+      if (!prev) return prev
```
steven-chiu-bigstack: Nitpick: It looks like `payload` will never be empty, so we can remove the `if (!prev) return prev` statement from all onChange handlers.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AttributeCheckboxGroup.tsx
```diff
@@ -0,0 +1,39 @@
...
+type AttributeCheckboxGroupProps = {
+  label: string
+  isAllChecked: boolean | null
+  attributes: string[]
+  selectedAttributes: string[]
+  onAttributesChange: (attribute: string) => void
+  onAllAttributesChange: () => void
+}
+
+export const AttributeCheckboxGroup = (props: AttributeCheckboxGroupProps) => {
```
steven-chiu-bigstack, line 13: The position of "All {attribute}" checkbox seems different from the mockup:

- Mockup

   image

- Implementation

   image

Both layouts look good to me, so either one is fine.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/AddDescription/AddDescription.tsx
```diff
@@ -0,0 +1,88 @@
...
+    errorMessage,
+    onNameChange,
+    onDescriptionChange,
+    onPublishClick: onPublishClickProp,
+  } = props
+
+  const isValueValid = useMemo(() => {
+    if (!payload.name) return false
+    return true
+  }, [payload.name])
```
steven-chiu-bigstack: Nitpick: `isValueValid` can be simplified to `!!payload.name` without `useMemo`, because checking whether a string is empty is fast.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AttributeResultPanel.tsx
```diff
@@ -0,0 +1,78 @@
...
+    predefinedEventIds,
+    onPanelClose,
+  } = props
+
+  const renderResultIds = () => {
+    if (isPredefinedEventsLoading)
+      return <CosSkeleton className="h-[16px] w-[120px]" />
+
+    if (predefinedEventIds.length === 0)
+      return <p className="primary-body4 text-functional-text">No Result</p>
```
steven-chiu-bigstack, line 37: I have a couple of questions about the implementation. I apologize if my comments seem nonsense, as I have not yet fully understand this feature! 🙇‍♂️

1. There's a text in the mockup says "When no attributes is added, shows “no result”". Does that mean a "No result" label should be displayed in the panel on the right when nothing is selected (on the left)?

   image

2. Continuing from 1, "no attributes is added" probably means there aren't any matching events to show. For this component (`AttributeResultPanel `), I think the following props interface would make more sense:

   ```ts
   type AttributeResultPanelProps = {
     isOpen: boolean
     isLoading: boolean
     matchingEventIds: string[] // <- Show "No result" as long as this prop is empty.
     onClose: () => void
   }
   ```

   And the parent component should filter event IDs based on the selected attributes and pass it down to `AttributeResultPanel`.

3. Continuing from 2, since there are no matching event IDs, this trigger will never activate once it's created. Therefore, the condition to proceed to the next step should be `matchingEventIds.length > 0`, rather than checking if any attribute is selected.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AddAttributeModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+    ),
+  }
+
+  const renderContent = activeType ? renderContentFnMap[activeType] : () => null
+
+  return (
+    <CosModal
+      isOpen={isModalOpen}
+      title="Add Attribute"
+      actionText="Set Response"
```
steven-chiu-bigstack: This should've been "Set Attributes" or "Add Attributes" right?

packages/cube-frontend-web-app/src/components/LogConsole.tsx
```diff
@@ -0,0 +1,68 @@
+import { twMerge } from 'tailwind-merge'
+import { CosHyperlink, CosTooltip } from '@cube-frontend/ui-library'
+
+export type LogConsoleProps = {
+  title: {
+    label: string
+    href?: string
```
raven-pan, line 7: It seems that `href` is not being used.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/FilePathCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+      false: 'cursor-pointer text-functional-text',
+    },
+  },
+})
+
+type FilePathCardProps = {
+  /**
+   * @default false
+   */
+  disabled?: boolean
```
raven-pan, line 29: It seems that `disabled` is not being used.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/useUploadScript.ts
```diff
@@ -0,0 +1,140 @@
...
+    })
+
+    try {
+      const testResult = await triggersApi.verifyTriggerScript({
+        dataCenter: dataCenter!.name,
+        verifyMaterialScriptRequest: { script: scriptInfo.content },
+      })
+
+      setShowScriptTestResult({
+        status: testResult.data.code === 200 ? 'testSucceeded' : 'testFailed',
```
raven-pan: If the code is not 200, it will be catch on `catch` block.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/PersonalizedScriptModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+            type="secondary"
+            usage="icon-left"
+            Icon={UploadIcon}
+            onClick={onUploadScriptButtonClick}
+          >
+            {`Upload ${scriptTypes?.types[0] ?? '-'} Script`}
+          </CosButton>
+          <input
+            ref={scriptFileInputRef}
+            type="file"
```
raven-pan: Discussion:
Maybe we can limit file types to ensure that the user always upload a shell script file.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AddAttributeModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+  const { alertTypes, severities, categories, eventIds } = attributes
+
+  const {
+    alertTypes: selectedAlertTypes,
+    severities: selectedSeverities,
+    categories: selectedCategories,
+    eventIds: selectedEventIds,
+  } = payload
+
+  const [tempAttributes, setTempAttributes] = useState<TriggerAttributes>({
```
steven-chiu-bigstack: The name "temp" is a bit confusing here. Also, properties like `payload.alertTypes` and `payload.severities` are non-primitive types, so it would be safer to deep clone them before putting them into a state to avoid issues caused by object reference. We could try something like the following:

```tsx
const {
  alertTypes: defaultSelectedAlertTypes,
  severities: defaultSelectedSeverities,
  categories: defaultSelectedCategories,
  eventIds: defaultSelectedEventIds,
} = payload

const cloneAttributes = useCallback((): TriggerAttributes => {
  return structuredClone({
    alertTypes: defaultSelectedAlertTypes,
    severities: defaultSelectedSeverities,
    categories: defaultSelectedCategories,
    eventIds: defaultSelectedEventIds,
  })
}, [
  defaultSelectedAlertTypes,
  defaultSelectedSeverities,
  defaultSelectedCategories,
  defaultSelectedEventIds,
])

const [selectedAttributes, setSelectedAttributes] = useState<TriggerAttributes>(cloneAttributes)

useEffect(() => {
  setSelectedAttributes(cloneAttributes())
}, [cloneAttributes])
```

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/useUploadScript.ts
```diff
@@ -0,0 +1,140 @@
...
+  const onFileChange: ChangeEventHandler<HTMLInputElement> = async (e) => {
+    const scriptFile = e.target.files?.[0]
+    if (!scriptFile) {
+      return
+    }
+
+    try {
+      const base64 = await fileToBase64(scriptFile)
+      setScriptInfo({
+        filePath: scriptFile.name,
```
steven-chiu-bigstack: While it's possible to retrieve a file path using `fileInputRef.current.value` once a file is chosen, browsers will obscure the actual path for security reasons. Considering this, do you think it make sense to rename `filePath` to `fileName`?

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/useUploadScript.ts
steven-chiu-bigstack: According to the API docs, `verifyTriggerScript` API will return 400 HTTP status code when the script is invalid, making this code block unreachable.

image

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/useUploadScript.ts
```diff
@@ -0,0 +1,140 @@
...
+        content: base64,
+      })
+      setShowScriptTestResult({
+        status: 'untested',
+        message: undefined,
+      })
+    } catch {
+      console.error('Failed to read script file.')
+    }
+  }
```
steven-chiu-bigstack, line 114: There's a common quirk with HTML file inputs where selecting the same file twice won't trigger the `onChange` event because the file path hasn't changed:

https://github.com/user-attachments/assets/f943ce40-19b1-42d2-b1b1-302b4070827a

To mitigate this issue, the most common solutions is to clear the input value when we're done working with it:

```ts
const input = fileInputRef.current

if (input) {
  input.value = ''
}
```

steven-chiu-bigstack (approved):
Hi @lizzy-liang-bigstack 

Thanks for the great work! I've left some comments for you to check out. Let me know if anything comes up. 🙏

packages/cube-frontend-web-app/src/components/UpsertTriggers/upsertTriggersUtils.ts
lizzy-liang-bigstack: It's actually a good point, thank you for pointing this out! 
I think the more intuitive expression would be:

```ts
return !isEmpty(emails) || !isEmpty(slacks) || !!script
```

packages/cube-frontend-web-app/src/components/LogConsole.tsx
lizzy-liang-bigstack: I was considering turning `LogConsole` into a shared component, similar to the one used on the Health Detail page and the future features. That’s why I included some optional props in the interface. My plan was to eventually replace the existing component with this shared version, if it makes sense to do so.

Maybe I should open a ticket to document this intention and track the change. 
What do you think?

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/PersonalizedScriptModal.tsx
lizzy-liang-bigstack: It's a fair point! This solution actually did come up during the discussion with the backend team. Bjergsen mentioned that the file type could vary depending on the scenario, and suggested encoding the file in Base64. The backend can then handle validation by checking whether it’s a valid script.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AttributeCheckboxGroup.tsx
lizzy-liang-bigstack: Ah, it's my bad! 
I've updated the implementation to better align with the design. 🙏

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AddAttributeModal.tsx
lizzy-liang-bigstack: Good point! I just noticed that the mockup has been updated, and "Add Attributes" is indeed what's suggested by the designer. Thanks for pointing it out! 🙂

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/useUploadScript.ts
lizzy-liang-bigstack: Sounds good!

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/UpsertTriggersSteps.tsx
lizzy-liang-bigstack: Good point! I’ll open a separate ticket to track the issues related to CosStepProcessItem, including the misuse you mentioned above, there’s a similar issue on the Tunings page as well.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/useUploadScript.ts
lizzy-liang-bigstack: Thanks for pointing that out!  I’ve added the suggestion code in the finally block of the `onFileChange` handler to ensure the input is cleared after processing.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AddAttributeModal.tsx
lizzy-liang-bigstack: Sounds good! I’ve updated the disable logic, now the button will be disabled if there are no changes from the previous selection.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AttributeResultPanel.tsx
lizzy-liang-bigstack: 1. Yes, the UI should display "No Result" when no attributes are selected. I'll remove the mock data to make it better align with the real use case.

2. Sounds good! I’ll update the props interface accordingly.

3. Actually, even if there are no matching event IDs, users should still be able to create a trigger with the selected attributes. This allows for future matches and gives users more flexibility. So I’d suggest we keep the current validation for the attributes field.

lizzy-liang-bigstack: https://github.com/user-attachments/assets/ca27deec-adde-4517-a4fc-a9afc6e3ef85

steven-chiu-bigstack: Got it, no problem! 👌 🙏

packages/cube-frontend-web-app/src/components/UpsertTriggers/useCreateTriggersPayload.ts
lizzy-liang-bigstack: Not really. I think that's just a stylistic choice 🤔

lizzy-liang-bigstack:
Two follow-up tickets have been created to handle related issues that raised from this PR:
- [[Enhancement] Make LogConsole a shared component for reuse across pages](https://github.com/bigstack-oss/cube-cos-ui/issues/578)
- [[Fix] Misuse of CosStepProcessItem and unclear prop naming](https://github.com/bigstack-oss/cube-cos-ui/issues/579)

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AttributeResultPanel.tsx
lizzy-liang-bigstack: Update:
Conclusion of 3. has been revised after checking with the designer: if there are no resulting event IDs, users will not be allowed to create triggers. Apologies for the earlier mistake.

PR #572, opened 2025-07-17
555 Edit Triggers Page

### LINKED ISSUE #555: [Feature] Edit Triggers
Enable support for editing existing triggers

1. As outlined in the [Mockup](https://www.figma.com/design/kSUzxXJ3OZ5RXSZCRjZ9o1/COS-3.1-M2-UI-mockup?node-id=12008-36110&m=dev).
2. Custom triggers should be fully editable.
3. Built-in triggers should support partial editing: the events and name fields shouldn't be not editable.

－

**Additional context**
－

lizzy-liang-bigstack:
⚠️ This PR is based on #570 and #571 

#### What type of PR is this?

- Feature

1. Select Events

https://github.com/user-attachments/assets/93ff407f-10b6-4558-8e1d-f3be04d81c86

2. Set Response and Add Description

https://github.com/user-attachments/assets/66819aff-0c11-46fa-b924-1642da89538a

Fixes #555 

> Note

#### Additional documentation

```docs

```


packages/cube-frontend-web-app/src/stores/editTriggersStore.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const createDefaultState = (): EditTriggersStoreState => ({
+  initialData: undefined,
+})
+
+type EditTriggersStoreActions = {
+  setInitialData: (initialData: GetTriggerResponseData) => void
+}
+
+export const useEditTriggersStore = create<
```
steven-chiu-bigstack: For the Edit Tunings feature, we created a store because tunings don't have unique IDs or names. Since we don't have a `getTuningDetail`-like API, this store will be essential for the Edit Tunings page to identify which tuning a user wants to edit.

However, for triggers, I think the [getTrigger](https://10.32.45.10/api/v1/datacenters/control/apidocs/index.html#/Triggers/getTrigger) API should work for fetching trigger details. Let me know if there's any issues! 🙏

packages/cube-frontend-ui-library/src/components/CosTextArea/CosTextArea.tsx
```diff
@@ -88,7 +88,7 @@ export const CosTextArea = (props: CosTextAreaProps) => {
           {label}
         </label>
         <div className="secondary-body4 text-functional-text-light">
-          {charCount}/{maxLength}
+          {charCount.toLocaleString()}/{maxLength.toLocaleString()}
```
steven-chiu-bigstack: We should explicitly define the `locales` for [Number.toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) to ensure the formatted string is always digits (e.g., using `(1234).toLocaleString('en-US')` for `1,234`)

For example (edge cases), on systems with Arabic locale settings, `(123456).toLocaleString()` returns `١٢٣٬٤٥٦`. Even though it is perfectly readable for native speakers, it significantly deviates from the mockup.

packages/cube-frontend-web-app/src/components/UpsertTriggers/upsertTriggersUtils.ts
```diff
@@ -96,3 +83,12 @@ export const shouldRedirectToListPage = (
...
   return false
 }
+
+export const filterEnumValues = <T extends string>(
+  rawValues: string[] | undefined,
+  enumObject: Record<string, T>,
+): T[] => {
+  return (rawValues ?? []).filter((value): value is T =>
+    Object.values(enumObject).includes(value as T),
+  )
```
steven-chiu-bigstack, line 94: Nitpick: This is minor, but we can make use of a `Set` to optimize the time complexity of this function from `M * N` to `M + N`:

```ts
const enumSet = new Set(Object.values(enumObject))
return (rawValues ?? []).filter((value) => enumSet.has(value as T))
```

packages/cube-frontend-web-app/src/components/UpsertTriggers/useCreateTriggersPayload.ts
```diff
@@ -61,14 +64,16 @@ export const useCreateTriggersPayload = (): UseCreateTriggersPayload => {
     }))
   }
 
-  const onCategorySelect = (categories: string[]): void => {
+  const onCategorySelect = (
+    categories: GetPredefinedEventsCategoriesEnum[],
+  ): void => {
```
steven-chiu-bigstack, line 69: (Note: The current implementation works well, so there's no need to change anything. 👍)
Sometimes a function creator could be useful for code deduplication. We have some similar cases in Events Page.

- Before (clear but repetitive)

   ```ts
   const onAlertTypeSelect = (
     alertTypes: GetPredefinedEventsTypesEnum[],
   ): void => {
     setPayload((prev) => ({
       ...prev,
       alertTypes,
     }))
   }
   
   const onSeveritySelect = (
     severities: GetPredefinedEventsSeveritiesEnum[],
   ): void => {
     setPayload((prev) => ({
       ...prev,
       severities,
     }))
   }
 
   const onCategorySelect = (
     categories: GetPredefinedEventsCategoriesEnum[],
   ): void => {
     setPayload((prev) => ({
       ...prev,
       categories,
     }))
   }

   // ...more change handlers with the same pattern.
   ```
- After (less code but also less straightforward)

   ```ts
   const createChangeHandler = <Key extends keyof UpsertTriggersPayload>(
     key: Key,
   ) => {
     return (value: UpsertTriggersPayload[Key]) => {
       setPayload((prev) => ({
         ...prev,
         [key]: value,
       }))
     }
   }

   const onAlertTypeSelect = createChangeHandler('alertTypes')
   const onSeveritySelect= createChangeHandler('severities')
   const onCategorySelect = createChangeHandler('categories')
   // ...more change handlers with the same pattern.
   ```

Straightforward code is usually better from my perspective (including this scenario), even if it means a few more lines.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/useStepParam.ts
```diff
@@ -30,24 +30,27 @@ export const useStepParam = (
   }, [searchParams, validSteps, validStepsSet])
 
   const goToSelectEvents = () => {
```
steven-chiu-bigstack: Nitpick: It seems that `goToSelectEvents` is never used. We can remove it from the codebase.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/useStepParam.ts
```diff
@@ -30,24 +30,27 @@ export const useStepParam = (
...
   const goToAddDescription = () => {
-    setSearchParams((prev) => ({
-      ...prev,
-      [STEP_PARAM_KEY]: UpsertTriggersStep.AddDescription,
-    }))
+    setSearchParams((prev) => {
+      const next = new URLSearchParams(prev)
+      next.set(STEP_PARAM_KEY, UpsertTriggersStep.AddDescription)
+      return next
+    })
```
steven-chiu-bigstack, line 44: Just curious, did you find anything wrong with the previous code? Or is this change just for optimization/readability?

packages/cube-frontend-web-app/src/pages/events/triggers/edit/EditTriggersPage.tsx
```diff
@@ -1,3 +1,82 @@
...
+import { TriggersApiGetTriggerRequest } from '@cube-frontend/api'
+import { UpsertTriggersPayload } from '@cube-frontend/web-app/components/UpsertTriggers/upsertTriggersUtils'
+import { DataCenterContext } from '@cube-frontend/web-app/context/DataCenterContext'
+import { useCosMutationRequest } from '@cube-frontend/web-app/hooks/useCosRequest/useCosMutationRequest'
+import { payloadToUpdateRequest } from '../utils'
+
 export const EditTriggersPage = () => {
-  return <div>Edit Triggers Page</div>
+  const navigate = useNavigate()
+  const [searchParams, _] = useSearchParams()
```
steven-chiu-bigstack: Nitpick: `, _` can be safely removed because `useSearchParams` returns an array.

packages/cube-frontend-web-app/src/pages/events/triggers/mockData.ts
steven-chiu-bigstack, line 1: It looks like this file is also not being used anywhere. Can we delete it?

packages/cube-frontend-web-app/src/components/UpsertTriggers/mockData.ts
steven-chiu-bigstack, line 1: It looks like this file is not being used anywhere. Can we delete it?

packages/cube-frontend-web-app/src/pages/events/triggers/edit/EditTriggersPage.tsx
```diff
@@ -1,3 +1,82 @@
...
+  const triggerName = searchParams.get('name')
+
+  if (!triggerName) {
+    throw new Error('Cannot find trigger name in the URL')
+  }
+
+  const { dataCenter } = useContext(DataCenterContext)
+
+  const [isUpdating, setIsUpdating] = useState(false)
```
steven-chiu-bigstack: We can replace `isUpdating` with `useCosMutationRequest().isLoading`.

packages/cube-frontend-web-app/src/pages/events/triggers/create/CreateTriggersPage.tsx
```diff
@@ -1,43 +1,43 @@
...
 
-  // const { dataCenter } = useContext(DataCenterContext)
+  const { dataCenter } = useContext(DataCenterContext)
 
-  // const {
-  //   mutateResource: createTrigger,
-  //   errorState,
-  //   clearError,
-  // } = useCosMutationRequest(triggersApi.createTrigger)
+  const [isCreating, setIsCreating] = useState(false)
```
steven-chiu-bigstack: We can replace `isCreating` with `useCosMutationRequest().isLoading`.

packages/cube-frontend-web-app/src/pages/events/triggers/_components/TriggersStatusToggle.tsx
```diff
@@ -13,18 +14,30 @@ export const TriggersStatusToggle = (props: TriggersStatusToggleProps) => {
...
 
-  const handleToggleChange = () => {
+  /**
+   * Prevents user interaction with the status toggle when:
+   * 1. The trigger has just been created, edited, or deleted
+   * 2. Until the trigger's `response` field has been successfully set
+   */
+  const isDisabled =
+    status?.current !== 'ok' ||
+    status?.isProcessing ||
```
steven-chiu-bigstack: It looks like we can remove line 27 and 28 as they are equivalent to `row.isProcessing`.

packages/cube-frontend-web-app/src/pages/events/triggers/_components/TriggersStatusToggle.tsx
```diff
@@ -13,18 +14,30 @@ export const TriggersStatusToggle = (props: TriggersStatusToggleProps) => {
...
+    isProcessing ||
+    types.length === 0
+
+  const handleToggleChange = useCallback(() => {
```
steven-chiu-bigstack: While `useCallback` isn't necessary here, it also doesn't cause any harm. 🤷‍♂️

packages/cube-frontend-web-app/src/pages/events/triggers/utils.ts
```diff
@@ -83,14 +83,25 @@ export const payloadToCreateRequest = (
...
-    // TODO: Align script request type better to API schema
-    script: {
-      filePath: payload.script!.fileName,
-      content: payload.script!.content,
-    },
-    notifications: {
-      emails: payload.emails.map((email) => email.address),
-      slacks: payload.slacks.map((slack) => slack.url),
-    },
+    script: payload.script ?? { name: '', content: '' },
```
steven-chiu-bigstack, line 86: To me, it feels kind of strange to pass a placeholder `script` object to the backend even if the user hasn't set any personalized script response. But it's just a DEV thing; users won't get impacted, so it's not a critical concern.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/PersonalizedScriptModal.tsx
```diff
@@ -92,7 +95,7 @@ export const PersonalizedScriptModal = (
               onChange={onFileChange}
             />
             <p className="primary-body2 text-functional-text">
-              {`OS: ${scriptTypes?.environments[0] ?? '-'}`}
+              {`OS: ${scriptTypes?.environment[0] ?? '-'}`}
```
steven-chiu-bigstack: The `[0]` in `language[0]` and `environment[0]` should be removed:

{E5588F31-00E8-4433-9708-97F006DD9214}

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/SetResponse.tsx
```diff
@@ -126,7 +126,7 @@ export const SetResponse = (props: SetResponseProps) => {
     return (
       <TriggersStackCard
         title="Personalized Script"
-        tags={[payload.script.fileName]}
+        tags={[payload.script.name]}
```
steven-chiu-bigstack, line 131: Notice a bug on the Edit Trigger page: if the edited trigger is created with a placeolder script `{ name: '', contnet: '' }`, the tag content in the stack card would be an empty string:

chrome_KZf4LRS4cO

We could check if both the `name` and `content` are empty. If yes, that probably means there's no script response set for this trigger.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/SetResponse.tsx
```diff
@@ -193,7 +189,7 @@ export const SetResponse = (props: SetResponseProps) => {
       <PersonalizedScriptModal
         isModalOpen={isPersonalizedScriptOpen}
         payload={payload}
-        scriptTypes={scriptTypes}
+        scriptTypes={scriptType}
```
steven-chiu-bigstack: Nitpick: This prop can be renamed to singular form.

packages/cube-frontend-web-app/src/components/UpsertTriggers/CreateTriggers.tsx
```diff
@@ -58,42 +53,55 @@ export const CreateTriggers = (props: CreateTriggersProps) => {
...
-  //     }),
-  //   )
+    return {
+      dataCenter: dataCenter!.name,
+      types: payload.alertTypes,
+      severities: payload.severities,
+      categories: payload.categories,
+      ids: payload.eventIds,
+    } satisfies EventsApiGetPredefinedEventsRequest
+  })
```
steven-chiu-bigstack: I haven't looked into this yet, but it looks like the `getPredefinedEvents` API would get called twice every time the attributes change, causing the UI to be unstable because there's no guarantee which response arrives first:

https://github.com/user-attachments/assets/72885443-61da-4f6a-ae34-a0a0ee7d7e6e

The bug happens in both Create Trigger and Edit Trigger pages.

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SelectEvents/AddAttributeModal.tsx
```diff
@@ -236,7 +242,9 @@ export const AddAttributeModal = (props: AddAttributeModalProps) => {
         isAllChecked={checkIsAllChecked(selectedAttributes.eventIds, eventIds)}
         attributes={eventIds}
         selectedAttributes={selectedAttributes.eventIds}
-        onAttributesChange={onSelectedEventIdChange}
+        onAttributesChange={
+          onSelectedEventIdChange as (attribute: string) => void
+        }
```
steven-chiu-bigstack: The type castings for all `onAttributesChange` seem unnecessary.

steven-chiu-bigstack (approved):
Hi @lizzy-liang-bigstack, thanks for taking care of this feature! 🙏  This one is quite challenging.

Here's some bugs I've found and a couple of suggestions:

1. It feels more natural to close the Personalized Script modal when the Set Response button is clicked.
   - By the way, it would be great to disable the "+ Personalized Script" button at the top if a script response is already set. This makes it clear that there can be only 1 script response per trigger. The current UI doesn't inform users about this restriction in any way.

      https://github.com/user-attachments/assets/e9bf635b-4313-488d-9503-18dc4e0ac029

2. When editing triggers with script responses, the modal will show the previously uploaded script name, but the same behavior is not applied to Create Trigger. This is a bit confusing.

   - Edit Trigger (script name is shown in the modal when the Edit button is clicked)

      image

      Apart from that, seeing a `LogConsole` with no messages is a bit odd as well. Do you think it makes sense to hide the console in this situation?

   - Create Trigger (script name not shown in the modal when the Edit button is clicked)

      chrome_MKdQCha5BM

3. We could show a not found UI or take the user back to the Events Triggers page if the edited trigger does not exist.

   chrome_ngU0wNjGhO

4. For the Personalized Script modal:
   1. I found it convenient if the modal could be automatically scrolled down to the bottom when the Test Running button is clicked & when the verification is done so that I can check the message without manual scrolling. It's possible that the message is so long that a scrollbar is inevitable, but the most important messages often show up at the bottom, so scrolling to the bottom is still a viable option.
   2. Showing the `data` from the `verifyTriggerScript` API could be useful for users as it shows the strings printed with `echo` in the script.
   chrome_PPk0zD03kx

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/useStepParam.ts
lizzy-liang-bigstack: I switched to using `URLSearchParams` because the previous implementation caused an issue where existing parameter like `name` in the URL was not properly preserved. 

There’s an official issue reporting a similar problem:
https://github.com/remix-run/react-router/issues/11449#issuecomment-2056650011

I think both ways below work, but I went with the second one for clarity.
What do you think?

# 1

```ts
    setSearchParams((prev) => ({
      ...Object.fromEntries(prev),
      [STEP_PARAM_KEY]: UpsertTriggersStep.SetResponse,
    }))
```

# 2

```ts
    setSearchParams((prev) => {
      const next = new URLSearchParams(prev)
      next.set(STEP_PARAM_KEY, UpsertTriggersStep.SetResponse)
      return next
    })
```

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/SetResponse.tsx
lizzy-liang-bigstack: Oops, my bad!
It should have been updated to match the API schema.
I've fixed it now 🙏

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/PersonalizedScriptModal.tsx
lizzy-liang-bigstack: Ahhhh, you're right!
It should align with the updated API schema. 
It's fixed now 🙇‍♀️

packages/cube-frontend-web-app/src/pages/events/triggers/mockData.ts
lizzy-liang-bigstack: No problem!

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/useStepParam.ts
steven-chiu-bigstack: Got it, thanks for the clarification! 👍 🚀
The latter (your current choice) should be safer according to the comments in the thread.

packages/cube-frontend-web-app/src/pages/events/triggers/edit/EditTriggersPage.tsx
```diff
@@ -13,17 +13,16 @@ import { payloadToUpdateRequest } from '../utils'
...
 export const EditTriggersPage = () => {
   const navigate = useNavigate()
-  const [searchParams, _] = useSearchParams()
+  const [searchParams] = useSearchParams()
   const triggerName = searchParams.get('name')
 
   if (!triggerName) {
-    throw new Error('Cannot find trigger name in the URL')
+    // throw new Error('Cannot find trigger name in the URL')
+    navigate('not-found', { replace: true })
```
steven-chiu-bigstack: It looks like this part is overlooked. 😅
Similar to `EditTriggers.tsx`, we can use `return <Navigate ...>` to take users back to Triggers page if `triggerName` is empty (and remove the comment 🙏)

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/PersonalizedScriptModal.tsx
```diff
@@ -36,33 +37,76 @@ export const PersonalizedScriptModal = (
...
+    ) as HTMLElement | null
+    if (!scrollableContainer) return
+
+    const isScrollable =
+      scrollableContainer.scrollHeight > scrollableContainer.clientHeight
+
+    if (isScrollable) {
+      target.scrollIntoView({ behavior: 'smooth', block: 'start' })
+    }
+  }, [isModalOpen, showScriptTestResult.message])
```
steven-chiu-bigstack, line 61: Regarding this effect:

1. To improve readability, we could either add a concise comment explaining what the effect does, or extract the logic into a custom hook with a descriptive name, something like `useScrollToLogConsoleOnMessageChange`.
2. Direct DOM manipulation such as `document.getElementById` is discouraged in React. A better pattern would be to use a ref to access the DOM element, as it keeps the interaction within React's lifecycle and management. We can add a prop `bodyRef?: Ref<HTMLDivElement>` to `CosModal` to allow its parent to interactive with it.
   - `scrollIntoView` can be called even if there's no scrollbar at all. No error will be thrown. But there's still no harm checking it beforehand.
   - The extra `log-console` container isn't necessary. After retrieving the scrollable element, we can scroll it to bottom using `scrollTo`:

      ```ts
      const modalBody = modalBodyRef.current

      if (modalBody) {
        modalBody.scrollTo({
          behavior: 'smooth',
          top: modalBody.scrollHeight,
        })
      }
      ```

packages/cube-frontend-web-app/src/components/LogConsole.tsx
```diff
@@ -60,7 +60,7 @@ export const LogConsole = (props: LogConsoleProps) => {
           </button>
         </CosTooltip>
       </div>
-      <div className="primary-body3 overflow-x-auto text-wrap rounded-b-[5px] bg-dark-700 px-6 py-4 text-functional-border-darker">
+      <div className="primary-body3 overflow-x-auto whitespace-break-spaces text-wrap rounded-b-[5px] bg-dark-700 px-6 py-4 text-functional-border-darker">
```
steven-chiu-bigstack: We have a similar case in Node IPMI page. `whitespace-pre` and `overflow-x-auto` is used because it's more friendly to log-like texts, especially for space-indented table.

packages/cube-frontend-web-app/src/components/UpsertTriggers/EditTriggers.tsx
```diff
@@ -114,6 +115,19 @@ export const EditTriggers = (props: EditTriggersProps) => {
...
+    const initialScriptContent = btoa(
+      initialData?.response.script.content ?? '',
+    )
+
+    return !(
+      initialData?.response.script.name === confirmedPayload.script?.name &&
+      initialScriptContent === confirmedPayload.script?.content &&
+      isEqual(initialData?.response.emails, confirmedPayload.emails) &&
+      isEqual(initialData?.response.slacks, confirmedPayload.slacks)
+    )
```
steven-chiu-bigstack, line 128: Nitpick: The combination of `!` and multiple `&&` could be difficult to understand sometimes. `!` and `every` is another way to do it:

```ts
return !every([
  isEqual(initialData?.response.script.name, confirmedPayload.script?.name),
  isEqual(initialScriptContent, confirmedPayload.script?.content),
  isEqual(initialData?.response.emails, confirmedPayload.emails),
  isEqual(initialData?.response.slacks, confirmedPayload.slacks)
])
```

steven-chiu-bigstack (approved):
Hi @lizzy-liang-bigstack 

Looks good, thanks! 🚀
I left a few more comments. It looks like the script name fix in COS has not been deployed yet ([discussion](https://bigstack-tech.slack.com/archives/C07GL6RKCHM/p1753947925866019))

{034F1017-D4D2-4EA4-B281-1778172FB12B}

However, I think that's out of frontend's scope. Feel free to merge this PR after frontend code issues are handled! 🙏

PR #608, opened 2025-08-12
fix: loading skeleton not shown during first load

raven-pan:
#### What type of PR is this?

fix

Loading skeleton not shown during first load.

Problem Screen Record:

https://github.com/user-attachments/assets/dde129aa-9b1f-4842-a7b3-41ff87f34cc9

Fixed Screen Record:

https://github.com/user-attachments/assets/6640c180-a18f-46d4-b307-16eebf7a9349

Root Cause:

Since the [React StrictMode](https://react.dev/reference/react/StrictMode) runs an extra setup+cleanup cycle for every Effect, the extra setup will executes `getResource` again and aborts previous fetch. Because the cancel error is caught in the  `next tick`, it overrides loading state to false.

Screenshot 2025-08-12 at 4 57 40 PM

image

Solution:

Since `cancel` is only executed on `unMount` or `refetch`, it's fine not to set the loading state to false.


steven-chiu-bigstack (approved):
LGTM, thank you so much for fixing this bug @raven-pan! 🙏

PR #614, opened 2025-08-15
610 notifications unread indicator

### LINKED ISSUE #610: [Feature] Notifications Page UI - Unread Indicator
Implement the unread indicator on the navbar and notification table rows.

- Unread notifications should only be marked as read when the user stays at the Notifications Page for more than 3 seconds.
- The notification icon in the navbar will display an indicator whenever there are unread notifications.

Implement the unread indicator on the navbar and notification table rows.A clear and concise description of what you want to happen.

None

**Additional context**

Image

<br />

Image

steven-chiu-bigstack:
#### What type of PR is this?

Feat

Implement the unread indicator for the function bar and the notification page table rows.

Fixes #610

https://github.com/user-attachments/assets/ace3ae96-3873-4cbf-9130-2f376aa176f9

Notifications should only be marked as read after staying at the Notifications Page for at least 3 seconds.

https://github.com/user-attachments/assets/2b39709a-2bd8-45cf-88de-eebae56eb5f5


packages/cube-frontend-web-app/src/pages/notifications/NotificationsPage.tsx
```diff
@@ -56,6 +72,21 @@ export const NotificationsPage = () => {
     return description
   }
 
+  const renderUnreadDot = (row: NotificationRow) => {
+    // Render nothing when all rows are read to prevent showing an empty, fixed-size span on every row.
+    if (!hasUnreadRows) return null
```
steven-chiu-bigstack, line 79: - Before (without this statement)

    image

- After (current implementation)

    image

packages/cube-frontend-web-app/src/pages/notifications/NotificationsPage.tsx
```diff
@@ -56,6 +72,21 @@ export const NotificationsPage = () => {
...
+
+    const isCreatedAfterLastAccessed = dayjs
+      .respectTzOffset(row.time)
+      .isAfter(lastAccessedAt)
+
+    return (
+      <span className="inline-flex size-4 items-center justify-center">
+        {isCreatedAfterLastAccessed && <UnreadNotificationDot />}
```
steven-chiu-bigstack, line 87: Conditionally shows the indicator instead of the entire container to avoid strange layout (to align the positions of all expand buttons):

image

packages/cube-frontend-ui-library/src/components/CosViewDetailsTable/CosViewDetailsTable.tsx
```diff
@@ -21,6 +22,7 @@ export type CosViewDetailsTableProps<ParentRow extends CosTableRow> = Omit<
     onExpandChange: (parentRowId: string, value: boolean) => void
     detailTitle?: string | ((parentRow: ParentRow) => string)
     getDetailItems: (parentRow: ParentRow) => CosViewDetailsTableDetailItem[]
+    beforeExpandButton?: (parentRow: ParentRow) => ReactNode
```
raven-pan, line 25: Discussion: If we add a use case demo in the Storybook, that would be awesome!

raven-pan (approved):
I noticed an issue:

There are some notifications from more then 24 hours ago:

https://github.com/user-attachments/assets/f6d1ab2d-65ac-41bb-a4e4-5b25d7d2a95c

Users can't see those notifications in first sight when they got to the Notification Page if they are older then 24 hours. However, the UI still marks them as read after 3 seconds, which can confuse user.

https://github.com/user-attachments/assets/8a4be2e7-2434-489c-b6a7-23fd8d222bc5

After this issue is resolved, feel free to merge this PR.
Thanks for your work!

packages/cube-frontend-ui-library/src/components/CosViewDetailsTable/CosViewDetailsTable.tsx
steven-chiu-bigstack: Right, I totally forgot about Storybook! 🙏

steven-chiu-bigstack:
> I noticed an issue:
> 
> There are some notifications from more then 24 hours ago:
> 
>  Screen.Recording.2025-08-16.at.5.23.09.PM.mov 
> Users can't see those notifications in first sight when they got to the Notification Page if they are older then 24 hours. However, the UI still marks them as read after 3 seconds, which can confuse user.
> 
>  Screen.Recording.2025-08-16.at.5.23.22.PM.mov 
> After this issue is resolved, feel free to merge this PR. Thanks for your work!

You're absolutely right, nice catch! 👍 
Based on our previous [discussion](https://www.figma.com/design/kSUzxXJ3OZ5RXSZCRjZ9o1?node-id=14620-17826#1372778442) with the team, notifications will only be kept for 30 days:

{3D142540-A683-4806-A718-AABD8AB2F61E}

I'll make the following adjustments to fix this issue:

1. If the last notification is older than 30 days, the unread indicator will remain **invisible**.
   - I'm not sure if the removal of notifications older than 30 days is already implemented; I'll confirm this with the backend developers. If it's guaranteed that the API won't ever return notifications older than 30 days, then this adjustment can be safely ignored.
2. Otherwise the default value in the time range dropdown will be automatically changed to make sure the last notification is always visible when entering the page.

Instead of reusing the last notification time from the function bar, we make a separate call to the getLastNotification API when a user enters the Notifications Page to avoid issues caused by the time gap in polling.

steven-chiu-bigstack:
Hi @raven-pan 

Thanks for your review and suggestion! I've changed the default selected time range from `24h` to `30d`, and added an example for the `beforeExpandButton` prop of `CosViewDetailsTable` in the Storybook:

{E11A9D4C-4765-4F76-8CB5-537309466176}

image

PR #624, opened 2025-08-19
feat: 623 show an empty screen before retrieving the user authentication status

### LINKED ISSUE #623: [Enhance] show an empty screen instead of the header and sidebar skeletons before retrieving the user authentication status
## Problem

The UI should display an empty screen (or apparently different loading strategy, like huge spinner) until the user's authentication status is confirmed, instead of showing the header and sidebar skeletons, since users may have security concerns.

https://github.com/user-attachments/assets/6abd40cb-c66f-4738-9051-53962d94051a

## Solution

The token is cookie-based, so the UI cannot access its expiration time in the browser.
We can only rely on the `userInfoApi.getMe` response to determine the user's authentication status.

raven-pan:
feat: 623 show an empty screen before retrieving the user authentication status

#### What type of PR is this?

feat

The UI should display an empty screen until the user's authentication status is confirmed,
instead of showing the header and sidebar skeletons, since users may have security concerns.

The token is cookie-based, so the UI cannot access its expiration time in the browser.
We can only rely on the `userInfoApi.getMe` response to determine the user's authentication status.

Fixes #623

#### Screen Recording

https://github.com/user-attachments/assets/318523ff-bd8f-40f3-9bfa-692392851f87

### DIFF
--- a/packages/cube-frontend-web-app/src/App.tsx
+++ b/packages/cube-frontend-web-app/src/App.tsx
@@ -19,7 +19,6 @@ import './tailwind.css'
 
 function App() {
   return (
-    // TODO: inject global css, setup global store here.
     <DataCenterProvider>
       <UserContextProvider>
         <ApplicationIntegrationsContextProvider>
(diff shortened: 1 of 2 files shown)


steven-chiu-bigstack (approved):
LGTM, thanks!

lizzy-liang-bigstack (approved):

PR #632, opened 2025-08-22
592 Handle No Data State in Dropdowns

### LINKED ISSUE #592: [Enhancement] Handle "No Data" State in Dropdown and SearchBar Dropdown
## Problem

Currently, both the `Dropdown` and `SearchBar Dropdown` components do not explicitly handle cases where there is no data to display.

## Proposed Solution

(Pending confirmation with designers)

Introduce a clear and consistent "No Data" fallback UI for both components.

lizzy-liang-bigstack:
#### What type of PR is this?

- Enhancement

##### 1. CosDropdown
- 1). When there is no data, display `No Data` in the menu ([77cc97f](https://github.com/bigstack-oss/cube-cos-ui/pull/632/commits/77cc97fa062883b57cf64851454d7420df853973))
- 2). Update the size of checkboxes in the dropdown to match the design guideline ([3ebea97](https://github.com/bigstack-oss/cube-cos-ui/pull/632/commits/3ebea976aea0b692d172598ce3cf4b7ba246cc99))

https://github.com/user-attachments/assets/4b24ac6c-5c65-4250-a18a-6837b932356a

Screenshot 2025-08-20 at 5 30 52 PM

| Screenshot 2025-08-20 at 5 32 00 PM |  Screenshot 2025-08-20 at 5 31 26 PM
|---------------------------|---------------------------|

##### 2. CosSearchBarFilter
- When there is no data, the menu should not open ([05e6408](https://github.com/bigstack-oss/cube-cos-ui/pull/632/commits/05e6408a59e9f9913393ef743bdd556ebf9bfa80))

https://github.com/user-attachments/assets/b5ae4b0e-c25d-4b75-9925-6521132c690f

##### 3. CosSearchBarGlobal
- 1). When there is no data, the menu should not open ([5373e00](https://github.com/bigstack-oss/cube-cos-ui/pull/632/commits/5373e007b1d20ffadca1c7a03a42e3a2b5003e27))
- 2). When there is no data, display `No Data` in the `Category` menu

https://github.com/user-attachments/assets/87d00ad1-1eb8-45df-b784-c6879782ee3a

Fixes #592 

> Note

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-ui-library/src/components/CosCheckbox/CosCheckbox.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosCheckbox/CosCheckbox.tsx
@@ -12,16 +12,22 @@ export type CosCheckboxColor =
   | 'secondary'
   | 'secondary-dark'
 
+export type CosCheckboxSize = 'md' | 'sm' | 'xs'
+
 export type CosCheckboxStatus = 'unselected' | 'selected' | 'indeterminate'
 
 export type CosCheckboxProps = Omit<
   InputHTMLAttributes<HTMLInputElement>,
-  'checked' | 'defaultChecked'
+  'checked' | 'defaultChecked' | 'size'
 > & {
   /**
    * @default primary
    */
   color?: CosCheckboxColor
+  /**
+   * @default md
+   */
+  size?: CosCheckboxSize
   label?: string
   labelClassName?: string
   /**
@@ -39,6 +45,7 @@ export type CosCheckboxProps = Omit<
 export const CosCheckbox = (props: CosCheckboxProps) => {
   const {
     color = 'primary',
+    size = 'md',
     label,
     labelClassName,
     id,
@@ -76,6 +83,12 @@ export const CosCheckbox = (props: CosCheckboxProps) => {
       return effectiveChecked ? CheckboxSelected : CheckboxUnselected
     })()
 
+    const iconSize = () => {
+      if (size === 'xs') return 'icon-sm'
+      if (size === 'sm') return 'icon-md-sm'
+      return 'icon-md'
+    }
+
     return (
       <div
         className={twMerge(
@@ -86,7 +99,7 @@ export const CosCheckbox = (props: CosCheckboxProps) => {
           }),
         )}
       >
-        <IconComponent className="icon-md" />
+        <IconComponent className={iconSize()} />
       </div>
     )
   }
@@ -94,7 +107,10 @@ export const CosCheckbox = (props: CosCheckboxProps) => {
   if (isLoading) return <CosCheckboxSkeleton />
 
   return (
-    <label htmlFor={id} className={twMerge(checkbox.container({ disabled }))}>
+    <label
+      htmlFor={id}
+      className={twMerge(checkbox.container({ size, disabled }))}
+    >
       <input
         {...restProps}
         id={id}
--- a/packages/cube-frontend-ui-library/src/components/CosCheckbox/CosCheckbox.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosCheckbox/CosCheckbox.tsx
@@ -12,23 +12,23 @@ export type CosCheckboxColor =
   | 'secondary'
   | 'secondary-dark'
 
-export type CosCheckboxSize = 'md' | 'sm' | 'xs'
+export type CosCheckboxLabelSize = 'md' | 'sm' | 'xs'
 
 export type CosCheckboxStatus = 'unselected' | 'selected' | 'indeterminate'
 
 export type CosCheckboxProps = Omit<
   InputHTMLAttributes<HTMLInputElement>,
-  'checked' | 'defaultChecked' | 'size'
+  'checked' | 'defaultChecked'
 > & {
   /**
    * @default primary
    */
   color?: CosCheckboxColor
+  label?: string
   /**
    * @default md
    */
-  size?: CosCheckboxSize
-  label?: string
+  labelSize?: CosCheckboxLabelSize
   labelClassName?: string
   /**
    * Use `null` for indeterminate state.
@@ -45,8 +45,8 @@ export type CosCheckboxProps = Omit<
 export const CosCheckbox = (props: CosCheckboxProps) => {
   const {
     color = 'primary',
-    size = 'md',
     label,
+    labelSize = 'md',
     labelClassName,
     id,
     defaultChecked = false,
@@ -83,12 +83,6 @@ export const CosCheckbox = (props: CosCheckboxProps) => {
       return effectiveChecked ? CheckboxSelected : CheckboxUnselected
     })()
 
-    const iconSize = () => {
-      if (size === 'xs') return 'icon-sm'
-      if (size === 'sm') return 'icon-md-sm'
-      return 'icon-md'
-    }
-
     return (
       <div
         className={twMerge(
@@ -99,7 +93,7 @@ export const CosCheckbox = (props: CosCheckboxProps) => {
           }),
         )}
       >
-        <IconComponent className={iconSize()} />
+        <IconComponent className="icon-md" />
       </div>
     )
   }
@@ -109,7 +103,7 @@ export const CosCheckbox = (props: CosCheckboxProps) => {
   return (
     <label
       htmlFor={id}
-      className={twMerge(checkbox.container({ size, disabled }))}
+      className={twMerge(checkbox.container({ size: labelSize, disabled }))}
     >
       <input
         {...restProps}
--- a/packages/cube-frontend-ui-library/src/components/CosDropdown/CosDropdownMenu.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosDropdown/CosDropdownMenu.tsx
@@ -67,13 +67,14 @@ export const CosDropdownMenu = (props: CosDropdownMenuProps) => {
       variant !== 'withFilter' || !searchValue
         ? allItems
         : allItems.filter((child) =>
-            String(child.props.children)
+            child.props.children
               .toLowerCase()
               .includes(searchValue.toLowerCase()),
           )
 
-    if (visibleItems.length === 0)
-      return <ItemNoData size={size} type={type} variant={variant} />
+    if (visibleItems.length === 0) {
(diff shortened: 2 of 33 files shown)


packages/cube-frontend-ui-library/src/components/CosDropdown/CosDropdownMenu.tsx
```diff
@@ -56,6 +58,27 @@ export const CosDropdownMenu = (props: CosDropdownMenuProps) => {
...
+  const renderMenuItems = () => {
+    const allItems = Children.toArray(children).filter(
+      checkIsCosDropdownItemElement,
+    )
+
+    const visibleItems =
+      variant !== 'withFilter' || !searchValue
+        ? allItems
+        : allItems.filter((child) =>
+            String(child.props.children)
```
steven-chiu-bigstack: It looks like `child.props.children` is already a string, so we should be able to omit the `String(...)` casting.

{86AB2136-DFD1-45E8-AA33-018F009C50F9}

packages/cube-frontend-ui-library/src/components/CosDropdown/_components/ItemCheckbox.tsx
```diff
@@ -35,6 +39,10 @@ export const ItemCheckbox = (props: ItemCheckboxProps) => {
     }
   }
 
+  const getSize = (): CosCheckboxLabelSize => {
```
steven-chiu-bigstack: Nitpick: We can rename this function to `getLabelSize`.

packages/cube-frontend-ui-library/src/components/CosSearchBar/CosSearchBarGlobal/GlobalSearchBarSorting.tsx
steven-chiu-bigstack, line 1: Not related to the implementation, just a quick rant.
I've always wondered why this dropdown is called "sorting." It doesn't actually sort anything. 🤔 

{817D19BD-8A56-472B-BEA2-1706523585CC}

To me, the purpose of the "category" here is to narrow down the search result, so something like `GlobalSearchCategoryDropdown` makes a lot more sense. But the mockups say "sorting," so I guess it is what it is. 🤷‍♂️

packages/cube-frontend-ui-library/src/components/CosSearchBar/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+  ReactNode,
+  ComponentProps,
+} from 'react'
+
+export const checkIsDropdownChildValid = <P, T extends ComponentType<P>>(
+  children: ReactNode,
+  expectedType: T,
+): children is
+  | ReactElement<ComponentProps<T>, T>
+  | ReactElement<ComponentProps<T>, T>[] => {
```
steven-chiu-bigstack: Nitpick: `ReactElement<ComponentProps<T>, T>` can be simplified as `ReactElement<P, T>`.

packages/cube-frontend-ui-library/src/stories/components/CosCheckbox/CosCheckbox.stories.tsx
```diff
@@ -36,9 +36,20 @@ export const Gallery: StoryObj = {
     return (
       <StoryLayout title="Checkbox">
         <StoryLayout.Section title="Checkbox">
-          <CheckboxGrid title="Master">
-            <CosCheckbox label={checkboxText} />
-          </CheckboxGrid>
+          <div className="flex flex-col gap-y-8">
```
steven-chiu-bigstack, line 39: Nitpick: It would be helpful to add a smaller title that says "Label Size" so people don't mistake it for the checkbox size.

packages/cube-frontend-ui-library/src/components/CosSearchBar/CosSearchBarFilter/CosSearchBarFilter.tsx
```diff
@@ -80,7 +85,7 @@ export const CosSearchBarFilter = (props: CosSearchBarFilterProps) => {
   }, [hasInputValue])
 
   const floatingProps = useFloating<HTMLDivElement, HTMLDivElement>({
-    isOpen: showDropdown && dropdownOpen,
+    isOpen: isValidChild && dropdownOpen,
```
steven-chiu-bigstack: Noticed a minor issue where the dropdown menu flashes when the input is cleared, which is caused by an `useEffect`:

https://github.com/user-attachments/assets/eafc04f0-d076-4a10-af1b-709581321599

{B18D7F78-41E5-4435-B4DF-E6FD766FF24E}

As a quick fix, we could remove the effect and adjust the `isOpen` value.

{70556D3E-D478-4205-A312-DDC193631783}

steven-chiu-bigstack (approved):
Hi @lizzy-liang-bigstack 

Nice work! I didn't look closely at the storybook-only code because it's not user facing. Let me know if there's anything you'd like to discuss about it. 🙏

packages/cube-frontend-ui-library/src/components/CosSearchBar/CosSearchBarFilter/CosSearchBarFilter.tsx
```diff
@@ -71,6 +71,11 @@ export const CosSearchBarFilter = (props: CosSearchBarFilterProps) => {
 
   const hasInputValue = value !== ''
 
+  const isValidChild = checkIsDropdownChildValid<
+    ComponentProps<typeof FilterDropdownItem>,
+    typeof FilterDropdownItem
+  >(children, FilterDropdownItem)
```
raven-pan: Discussion: How about renaming it to `hasMenuItems` or `hasMenuOptions`?

packages/cube-frontend-ui-library/src/components/CosCheckbox/CosCheckbox.tsx
```diff
@@ -12,17 +12,23 @@ export type CosCheckboxColor =
...
   | 'secondary-dark'
 
+export type CosCheckboxLabelSize = 'md' | 'sm' | 'xs'
+
 export type CosCheckboxStatus = 'unselected' | 'selected' | 'indeterminate'
 
 export type CosCheckboxProps = Omit<
   InputHTMLAttributes<HTMLInputElement>,
-  'checked' | 'defaultChecked'
+  'checked' | 'defaultChecked' | 'size'
```
raven-pan: Question: Why do we ignore the `size` prop here?

packages/cube-frontend-ui-library/src/components/CosSearchBar/utils.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  Children,
+  ComponentType,
+  isValidElement,
+  ReactElement,
+  ReactNode,
+  ComponentProps,
+} from 'react'
+
+export const checkIsDropdownChildValid = <P, T extends ComponentType<P>>(
```
raven-pan: Discussion: How about defining the type like this:

```tsx
export const checkIsDropdownChildValid = <P>(
  children: ReactNode,
  expectedType: ComponentType<P>,
): children is ReactElement<P> | ReactElement<P>[] => {
```

And we can just let TypeScript derive the <P> from the second parameter:

```tsx
const isValidChild = checkIsDropdownChildValid(
  children,
  CosSearchBarGlobalItem,
)
```

packages/cube-frontend-ui-library/src/components/CosCheckbox/CosCheckbox.tsx
```diff
@@ -12,17 +12,23 @@ export type CosCheckboxColor =
...
 > & {
   /**
    * @default primary
    */
   color?: CosCheckboxColor
+  /**
```
raven-pan, line 28: It should placed before labelSize prop.

raven-pan (approved):

packages/cube-frontend-ui-library/src/components/CosCheckbox/CosCheckbox.tsx
lizzy-liang-bigstack: Ah, you're right! Since we’re using labelSize instead of size, we don’t need to omit the size prop here. Thanks for pointing this out :)

packages/cube-frontend-ui-library/src/components/CosSearchBar/CosSearchBarFilter/CosSearchBarFilter.tsx
lizzy-liang-bigstack: Sounds good!

packages/cube-frontend-ui-library/src/components/CosSearchBar/CosSearchBarGlobal/GlobalSearchBarSorting.tsx
lizzy-liang-bigstack: I totally agree! I can check with the designers to see if we can change the naming from `sorting` to something like `category`, which would definitely make more sense. I'll update it in the next commit, thanks for bringing this up!

packages/cube-frontend-ui-library/src/components/CosSearchBar/utils.ts
lizzy-liang-bigstack: That’s indeed a cleaner way to define the type, great suggestion!

packages/cube-frontend-ui-library/src/stories/components/CosCheckbox/CosCheckbox.stories.tsx
lizzy-liang-bigstack: No problem! I’ve updated the Storybook, hopefully this makes it clearer for users.

Screenshot 2025-09-01 at 4 43 26 PM

packages/cube-frontend-ui-library/src/components/CosSearchBar/CosSearchBarGlobal/GlobalSearchBarSorting.tsx
lizzy-liang-bigstack: Update: the designer agreed on renaming `sorting` to `category` for better align with its usage.

Issue #503, opened 2025-06-04
UI Library - Panel components should be applied to all COS UI Pages


raven-pan:
## Problem

COS Design Guideline & Storybook:

<img width="1908" alt="Image" src="https://github.com/user-attachments/assets/730972d0-fef8-4c43-a1f7-05eeb3cc1138" />

Current COS 3.0.0 UI:

<img width="1491" alt="Image" src="https://github.com/user-attachments/assets/7280cce6-ae39-4f86-9ba1-95dd4555c9fd" />

<img width="1490" alt="Image" src="https://github.com/user-attachments/assets/ecb6da24-bb79-4ec7-a088-7bbb88efb547" />

Which page is not using  the Panel components?

1. Health, Health Details
2. Events
3. Tuning
4. Trigger
5. Settings

## Notes

Discuss with the designers first, since some mockups do not use the Panel component.

PR #635, opened 2025-08-22
634 Update Redirect Search Param Keys and Refactor Events Page

### LINKED ISSUE #634: [Bug] Missing Search Params When Redirecting from Chart Page to Events Page
## Problem

When redirecting from the Chart page to the Events page by clicking a pie or bar chart component, the search params were not being passed correctly. This causes filters and query states to be lost, causing inconsistent behaviour between the two pages.

https://github.com/user-attachments/assets/d50f4fa8-9bc2-4420-94f6-88507ced33f6

lizzy-liang-bigstack:
#### What type of PR is this?

- Fix

1. Update search param keys when redirecting from the Chart page to the Events page to make sure filters are correctly set ([94fbd59](https://github.com/bigstack-oss/cube-cos-ui/pull/635/commits/94fbd59b5930c17f65d2fea532f37987045150f0))

https://github.com/user-attachments/assets/0c7afa22-f618-432b-b539-4481a9acf18a

2. Align query keys with the API schema for consistency ([be08f17](https://github.com/bigstack-oss/cube-cos-ui/pull/635/commits/be08f17f482f22695a5a3e29d40dbc0885f119da))

3. Flatten the folder structure of the Events module for the better maintainability ([1291c61](https://github.com/bigstack-oss/cube-cos-ui/pull/635/commits/1291c6144c4e1f77903e7e236c09dec59a538478))

4. Refactor hooks by reusing the shared `useEventsFilter` hook and removing the duplicate implementation ([1291c61](https://github.com/bigstack-oss/cube-cos-ui/pull/635/commits/1291c6144c4e1f77903e7e236c09dec59a538478))

Fixes #634 

> Note

#### Additional documentation

```docs

```


packages/cube-frontend-web-app/src/pages/events/chart/_components/utils.ts
```diff
@@ -176,20 +176,13 @@ export const getRedirectUrl = (
...
-    keyword: id,
-    start: startDate.format(),
-    stop: endDate.format(),
+    keyword: targetEvent.id,
+    startDate: startDate.format(),
+    endDate: endDate.format(),
   })
 
-  if (category) searchParams.set('category', category)
-  if (severity) searchParams.set('severity', severity)
```
steven-chiu-bigstack, line 189: I noticed that `category` and `severity` were omitted in the new implementation. I think it's fine as long as the `keyword` alone is enough to narrow and get the right search results. For example, `NET00001I` seems to already include both the `NET` category and `INFO` severity.

packages/cube-frontend-web-app/src/pages/events/index/EventsIndexPage.tsx
```diff
@@ -1,5 +1,56 @@
...
+
+  const {
+    events,
+    isEventsLoading,
+    getResource: onEventsRefresh,
+    totalItems,
+  } = useEvents(eventsQuery)
+
+  return (
+    <div className="flex flex-col gap-6 bg-white px-6 py-4">
```
steven-chiu-bigstack, line 28: Minor question: Do you think it's suitable to use `CosGeneralPanel` here? It seems to be the default container we've been using when implementing new pages recently.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTableFilter.tsx
```diff
@@ -121,10 +118,10 @@ export const EventsTableFilter = (props: EventsTableFilterProps) => {
...
     stop: endDate,
-    category,
-    severity,
-    host,
-    instance,
+    categories,
+    severities,
+    hosts,
+    instances,
   }).some((value) => !!value)
```
steven-chiu-bigstack: Notice a minor issue where the clear filter X button is always visible because params such as `categories`, `severities`, etc. are always present as an array:

image

We can change `!!value` to `isEmpty(value)` (from lodash) as a quick fix.

packages/cube-frontend-web-app/src/pages/events/index/_components/utils.ts
```diff
@@ -136,42 +136,40 @@ export const queryToSearchParams = (query: EventsQuery): URLSearchParams => {
...
+  nextSearchParams.set(EventsParamKeyEnum.ItemsPerPage, itemsPerPage.toString())
+
   if (keyword) {
     nextSearchParams.set(EventsParamKeyEnum.Keyword, keyword)
   }
 
-  category.forEach((category) => {
-    nextSearchParams.append(EventsParamKeyEnum.Category, category)
-  })
+  if (categories?.length !== 0)
```
steven-chiu-bigstack: Nitpick: We should replace `if (categories?.length !== 0)` with `if (categories?.length)` or `if (!isEmpty(categories))` because `categories?.length` would resolve to `undefined` when `categories` is `undefined`. The current code runs because `categories` will never be `undefined`.

Or we can just run the for loop no matter what. It's okay to skip the null & length check:

- Before

    ```ts
    if (categories?.length) {
      categories.forEach((category) => ...)
    }
    ```

- After

    ```ts
    categories?.forEach((category) => ...)
    ```

steven-chiu-bigstack (approved):
Hi @lizzy-liang-bigstack 

LGTM, thanks for cleaning things up! 👍 🙏

packages/cube-frontend-web-app/src/pages/events/index/EventsIndexPage.tsx
lizzy-liang-bigstack: Yes, I completely agree!
In fact, we already have another ticket tracking this issue (#503).
The related PR should be opened soon, thanks for pointing it out! 🙂

packages/cube-frontend-web-app/src/pages/events/chart/_components/utils.ts
lizzy-liang-bigstack: Yes, I felt the keyword needed to be more precise for searching events.

packages/cube-frontend-web-app/src/pages/events/index/_components/EventsTableFilter.tsx
lizzy-liang-bigstack: Ah, that’s a great point! I’ll update it in the next commit.

packages/cube-frontend-web-app/src/pages/events/index/_components/utils.ts
lizzy-liang-bigstack: Sounds good!
Thanks a lot for the helpful suggestion!

lizzy-liang-bigstack:
Hi @bigstack-oss/qa,

The latest UI RPM can be found here:
https://github.com/bigstack-oss/cube-cos-ui/releases/tag/cube-cos-ui-3.1.0-dev-564c496

PR #641, opened 2025-08-25
612 Create CosUpload Component

### LINKED ISSUE #612: [UI library]: Upload Component
[Mockup](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=9614-11894&m=dev)

lizzy-liang-bigstack:
#### What type of PR is this?

- Feat

1. Create `CosUpload` component along with its Storybook entry ([0d011c4](https://github.com/bigstack-oss/cube-cos-ui/commit/0d011c48e066ddd13dfbed701c5bdef44653d5e1))

Screenshot 2025-08-25 at 4 57 09 PM

https://github.com/user-attachments/assets/ee9b902f-0b74-4d5e-b98e-1ded568bc0ff

2. Replace existing custom upload components in the web-app  with `CosUpload`, the behaviour remains unchanged ([da7e286](https://github.com/bigstack-oss/cube-cos-ui/commit/da7e2864c5362b34d18317001af97aee0ce3ddd9))

https://github.com/user-attachments/assets/89447059-8ab7-499b-9e38-273d0d3c262a

Fixes #612 

> Web-app: The `PersonalizedScriptModal` locates in `/events/triggers/create` and `/events/triggers/edit`)

#### Additional documentation

```docs

```


packages/cube-frontend-ui-library/src/components/CosUpload/CosUpload.tsx
```diff
@@ -0,0 +1,59 @@
...
+export type CosUploadProps = {
+  disabled: boolean
+  button: React.ReactElement<CosButtonProps>
+  input: React.ReactElement<HTMLInputElement>
+
+  leftSlot?: React.ReactNode
+  rightSlot?: React.ReactNode
+  children?: React.ReactNode
+  errorMessage?: string
+}
```
steven-chiu-bigstack, line 16: Hi @lizzy-liang-bigstack 

Looks good! Quick question—what do you think of the following props interface? I feel like it would make the component easier to use:

```ts
type CosUploadProps = {
  buttonText: string
  inputId?: string
  accept?: string
  leftSlot?: ReactNode
  rightSlot?: ReactNode
  errorMessage?: string
  disabled?: boolean
}
```

This ensures the button has a consistent style and icon, and it also embeds the hidden input element inside, with a customizable `inputId` to link with labels (e.g., `<label for="my-input-id">`).

On a separate note, the designer and I discussed the firmware/fixpack upload UI yesterday. Some changes will be made to the Upload component soon, though the mockup hasn't been updated yet. Some of the main adjustments are:

1. During upload, the button will be in a loading state instead of just disabled.
2. Adding an optional progress bar for large files (not for upserting triggers because the upload is fast there.)
3. Replacing the error message with `<CosInlineNotification type="error" title="reason" />`.

Since these changes are tied to the updated mockup, let's exclude them from this PR for now and stick to the current design.

By the way, it would be great if we can add an example for the `errorMessage` prop in the story, thanks! 🙏

steven-chiu-bigstack: Update: I've added one more prop `accept?: string` to limit the file extension. 😬

lizzy-liang-bigstack: Hi @steven-chiu-bigstack,

You’re absolutely right, this change does make the component easier to use!
I’ve updated the props interface based on your suggestion in the latest commit.
Please let me know if you have any further questions or feedback 🙂

packages/cube-frontend-ui-library/src/components/CosUpload/CosUpload.tsx
```diff
@@ -0,0 +1,96 @@
...
+    children,
+    onFileChange,
+  } = props
+
+  const fileInputRef = useRef<HTMLInputElement>(null)
+
+  const [internalError, setInternalError] = useState<string | null>(null)
+  const errorMessage = errorMessageProp || internalError
+
+  const defaultId = useId()
```
steven-chiu-bigstack: Nitpick: I think default id is not necessary, but it's causing no harm. 👌

packages/cube-frontend-ui-library/src/components/CosUpload/CosUpload.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const handleFileChange: ChangeEventHandler<HTMLInputElement> = async (e) => {
+    setInternalError(null)
+    const file = e.target.files?.[0] || null
+
+    if (onFileChange) {
+      try {
+        await onFileChange(file)
+      } catch (error) {
+        console.error(error)
+        setInternalError('Something went wrong, please try again.')
```
steven-chiu-bigstack: Since we have an `errorMessage` prop, it feels more natural to me to leave the error handling to the user. Users should try/catch the `onFileChange` function when necessary and update the `errorMessage` prop accordingly. This helps keep the `CosUpload` component simple without introducing any internal states.

packages/cube-frontend-ui-library/src/components/CosUpload/CosUpload.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const defaultId = useId()
+  const inputId = inputIdProp || defaultId
+
+  const handleButtonClick = () => {
+    fileInputRef.current?.click()
+  }
+
+  const handleFileChange: ChangeEventHandler<HTMLInputElement> = async (e) => {
+    setInternalError(null)
+    const file = e.target.files?.[0] || null
```
steven-chiu-bigstack, line 38: Side note: HTML file inputs actually support selecting multiple files with the [multiple](https://www.w3schools.com/tags/att_input_multiple.asp) attribute. That's where `e.target.files` gets its name.
But it seems that `CosUpload` is designed for single file upload at the current stage, so your implementation looks great! 👍

steven-chiu-bigstack (approved):
Hi @lizzy-liang-bigstack 

Nice work! I've left some more comments for you to check. Feel free to merge this PR when you're ready! 🙏 🚀 💯

packages/cube-frontend-ui-library/src/components/CosUpload/CosUpload.tsx
lizzy-liang-bigstack: Actually, I've thought about it, and as you mentioned, the current `CosUpload` only supports a single file. 
That said, it could definitely be extended in the future if needed. 💃
Thanks again for pointing this out!

packages/cube-frontend-ui-library/src/components/CosUpload/CosUpload.tsx
lizzy-liang-bigstack: That totally makes sense! I agree that showing error message from `errorMessage` prop sounds like a cleaner approach. I’ll adjust the implementation accordingly 👍

PR #642, opened 2025-08-26
639 Update Button Loading State's Styles

### LINKED ISSUE #639: [UI Enhancement] Button Loading State
[Mockup](https://www.figma.com/design/T95ymFSTSDKdmr9WEjylQE/COS-Design-System-2024?node-id=41-282&m=dev)

lizzy-liang-bigstack:
#### What type of PR is this?

- Feat

- Update Button Loading State's Styles

https://github.com/user-attachments/assets/2deb0c0a-e8ba-4fdd-b7df-d5de9eae6f3d

Fixes #639 

> Note

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-ui-library/src/components/CosButton/cosButtonStyles.tsx
+++ b/packages/cube-frontend-ui-library/src/components/CosButton/cosButtonStyles.tsx
@@ -1,0 +1,234 @@
+import { cva } from 'class-variance-authority'
+import { CosButtonUsage } from './CosButton'
+import { ClassValue } from 'class-variance-authority/types'
+
+export const button = cva(
+  'flex shrink-0 items-center justify-center gap-x-2 rounded-[5px] font-urbanist font-semibold transition-colors',
+  {
+    variants: {
+      type: {
+        primary: [
+          'bg-primary text-grey-0',
+          'hover:bg-functional-hover-primary',
+        ],
+        secondary: [
+          'border border-primary bg-grey-0 text-primary',
+          'hover:border-functional-hover-primary hover:bg-functional-hover-secondary',
+        ],
+        ghost: [
+          'bg-transparent text-primary',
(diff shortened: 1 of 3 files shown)


packages/cube-frontend-ui-library/src/components/CosButton/cosButtonStyles.tsx
```diff
@@ -0,0 +1,207 @@
...
+        md: 'secondary-body2 h-[34px] px-4 py-2',
+        lg: 'secondary-body2 h-[42px] px-5 py-3',
+      },
+      usage: {} as Record<NonNullable<CosButtonUsage>, ClassValue>,
+      loading: {
+        true: 'pointer-events-none cursor-default',
+        false: '',
+      },
+      disabled: {
+        true: 'pointer-events-none cursor-default',
```
steven-chiu-bigstack: Is there any cases where `pointer-events-none` is necessary for buttons in a loading or disabled state? If not, I suggest we remove this class for better UX. On the Maintenance Update page, we have a table where both the rows and the buttons inside them are clickable. However, buttons can be disabled under certain circumstances:

{3359C358-4E88-4581-8F31-B40988C4E316}

The problem with `pointer-events-none` is that when a user tries to click a disabled button in the table, the click event will pass through and trigger the row's click action instead.

steven-chiu-bigstack (approved):
LGTM, thanks! 🙏

packages/cube-frontend-ui-library/src/components/CosButton/cosButtonStyles.tsx
lizzy-liang-bigstack: It's a fair point to remove `pointer-events-none` since we don't want to pass through the click event accidentally. The class was initially used to disable the hover styles, I've added the hover styles manually instead, thanks for pointing this out! :)

PR #658, opened 2025-09-05
Fix: Display Verify Script Message Correctly

lizzy-liang-bigstack:
#### What type of PR is this?

- fix

- When verifying a personalised script during trigger creation or editing, if the verification fails, display the corresponding status and message from the response.

### Before

https://github.com/user-attachments/assets/bf2ba26a-d47f-44ed-aba3-b3ae3c27221b

### After

https://github.com/user-attachments/assets/69bbfea5-afd7-449b-acfe-6547e8e38d3c

Fixes #

> Note

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/useUploadScript.ts
+++ b/packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/useUploadScript.ts
@@ -1,4 +1,6 @@
 import { useContext, useEffect, useState } from 'react'
+import { isAxiosError } from 'axios'
+import { upperFirst } from 'lodash'
 import { DataCenterContext } from '@cube-frontend/web-app/context/DataCenterContext'
 import { fileToBase64 } from '@cube-frontend/web-app/utils/file'
 import { triggersApi } from '@cube-frontend/web-app/api/cosApi'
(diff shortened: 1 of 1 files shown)


packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/useUploadScript.ts
```diff
@@ -107,17 +109,31 @@ export const useUploadScript = (
         verifyMaterialScriptRequest: { script: scriptInfo.content },
       })
 
-      const formattedResult = `Script:\n${testResult.data.data}\n\nResult:\n${testResult.data.msg}`
+      const formattedResult =
+        `Script:\n` +
+        `${testResult.data.data}\n\n` +
+        `Result:\n${upperFirst(testResult.data.msg)}`
```
steven-chiu-bigstack, line 115: Note: no need to make any changes, just wanted to mention that the `join` method can help us avoid manually appending `\n` to each line:

```ts
const formattedResult = [
  'Script:',
  `${testResult.data.data}\n`,
  `Result:\n${upperFirst(testResult.data.msg)}`,
].join('\n')
```

packages/cube-frontend-web-app/src/components/UpsertTriggers/_components/SetResponse/PersonalizedScriptModal/useUploadScript.ts
```diff
@@ -107,17 +109,31 @@ export const useUploadScript = (
...
       setShowScriptTestResult({
         status: 'testSucceeded',
         message: formattedResult,
       })
     } catch (error) {
-      setShowScriptTestResult({
-        status: 'testFailed',
-        message: 'Error occurred when verifying script: ' + error,
-      })
+      if (error instanceof AxiosError) {
```
steven-chiu-bigstack: Note: this works, but Axios also comes with a built-in method for checking error type:

```ts
import { isAxiosError } from 'axios'

if (isAxiosError(error)) {
  // Do something with `error`.
}
```

Still, there's no need to make any changes as it's very subtle.

steven-chiu-bigstack (approved):
LGMT, thanks @lizzy-liang-bigstack! 👍 🙏

PR #659, opened 2025-09-06
645 maintenance update page upload fixpack

### LINKED ISSUE #645: [Feature] Maintenance Update Page Upload Fixpack
Allow users to upload new fixpack PKG files and perform an MD5 checksum verification.

Allow users to upload new fixpack PKG files and perform an MD5 checksum verification.

None

**Additional context**
[Mockup](https://www.figma.com/design/kSUzxXJ3OZ5RXSZCRjZ9o1/COS-3.1-M2-UI-mockup?node-id=14882-17014&m=dev)

Image

steven-chiu-bigstack:
#### What type of PR is this?

Feat

- Implement upload fixpack feature on Maintenance Update page.
- Implement delete fixpack feature.

Fixes #645

1. The delete fixpack API still needs some adjustments, and is currently not working.
   - Therefore, to test the upload feature a few more times, the following approach should be used to delete uploaded fixpacks before the delete fixpack API is fixed:
      ```bash
      > ssh root@10.32.45.10
      > cd /var/appliance-db && rm fixpack.history
      > cd /var/fixpack && rm *.fixpack
      ```
2. Fixpack ID is embedded in the file's content. The API server automatically sorts the fixpacks based on a predefined order in COS (e.g., FIX_003 before FIX_002). This means the upload order doesn't matter, as long as the fixpacks aren't installed yet.
   - Uploading a fixpack with an ID that is already uploaded will result in a 409 Conflict response.
3. In the mockup, a fixpack is added to the list only after a user clicks the "Next" button. However, in the current implementation (both in frontend and backend), the fixpack is added to the list as soon as the MD5 verification is completed successfully.  I've asked designers if they're okay with this ~but haven't heard back yet. I'll make an update here when I get more information.~
   - 🆕Update: The designer has agreed to proceed with the current implementation due to the deadline, and developers will make adjustments when necessary if there's enough time.
   image

#### Additional documentation

- Fixpacks and checksum files for testing: [fixpacks.zip](https://github.com/user-attachments/files/22184783/fixpacks.zip)

- Upload succeeded

   https://github.com/user-attachments/assets/f6406874-3c58-42ab-ad1c-ae6867d6974f

- Confirm before closing the modal when a file is being uploaded / is uploaded.

   https://github.com/user-attachments/assets/d62d03d1-89d2-4df9-b8a8-5163580fb8f4

- Delete fixpack

   https://github.com/user-attachments/assets/6d2d5e41-84bd-406f-af73-adb3bcb789a7

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -40,6 +40,7 @@
     "fixpacks",
     "rollbackable",
     "unrollbackable",
-    "rollbackability"
+    "rollbackability",
+    "hoverable"
   ]
 }
--- a/packages/cube-frontend-api/sdk/api.ts
+++ b/packages/cube-frontend-api/sdk/api.ts
@@ -1173,31 +1173,6 @@ export interface DeleteFixpack200Response {
 /**
  * 
  * @export
- * @interface DeleteFixpack404Response
- */
-export interface DeleteFixpack404Response {
-    /**
-     * 
-     * @type {number}
-     * @memberof DeleteFixpack404Response
-     */
-    'code': number;
-    /**
-     * 
-     * @type {string}
-     * @memberof DeleteFixpack404Response
-     */
-    'msg': string;
-    /**
-     * 
-     * @type {string}
-     * @memberof DeleteFixpack404Response
-     */
-    'status': string;
-}
-/**
- * 
- * @export
  * @interface DeleteFixpack500Response
  */
 export interface DeleteFixpack500Response {
@@ -1608,56 +1583,6 @@ export interface DiskReadWriteHistory {
      * @memberof DiskReadWriteHistory
      */
     'write': Array<MetricRankRankInnerHistoryInner>;
-}
-/**
- * 
- * @export
- * @interface DrainNode200Response
- */
-export interface DrainNode200Response {
-    /**
-     * 
-     * @type {number}
-     * @memberof DrainNode200Response
-     */
-    'code'?: number;
-    /**
-     * 
-     * @type {string}
-     * @memberof DrainNode200Response
-     */
-    'msg'?: string;
-    /**
-     * 
-     * @type {string}
-     * @memberof DrainNode200Response
-     */
-    'status'?: string;
-}
-/**
- * 
- * @export
- * @interface DrainNode500Response
- */
-export interface DrainNode500Response {
-    /**
-     * 
-     * @type {number}
-     * @memberof DrainNode500Response
-     */
-    'code'?: number;
-    /**
-     * 
-     * @type {string}
-     * @memberof DrainNode500Response
-     */
-    'msg'?: string;
-    /**
-     * 
-     * @type {string}
-     * @memberof DrainNode500Response
-     */
-    'status'?: string;
 }
 /**
  * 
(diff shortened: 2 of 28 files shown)


raven-pan (approved):
Great works! Just a style issue and some discussion.
Feel free to merge. Thanks!

1.) There's a style issue:

Image

The border radiuses is not correct, but this is just a minor style issue.

2.) Discussion

I think the current verify API design may encounter some edge cases, for example:

1. If 2 members upload different fixpacks at the same time
2. member_1  uploads the FIXPACK_1 file, then uploads FIXPACK_1’s md5.
3. member_2 uploads the FIXPACK_2 md5, then uploads FIXPACK_2’s md5.

What will happen?

3.) Just a curios question 🙏, why we need to send file name to the API?

Image

raven-pan (commented):
UI/UX Discussion:

I saw you implemented the `browser confirm` to remind the user that the file upload isn't complete yet when they manually close the modal, it's great and very user friendly.

Maybe we could also implement this hint when the user tries to close the browser tab, if the fixpack/firmware file is large.

Just a minor discussion, so we don't need to implement this at this phase.

steven-chiu-bigstack:
> Great works! Just a style issue and some discussion. Feel free to merge. Thanks!
> 
> 1.) There's a style issue:
> 
> Image
> The border radiuses is not correct, but this is just a minor style issue.
> 
> 2.) Discussion
> 
> I think the current verify API design may encounter some edge cases, for example:
> 
> 1. If 2 members upload different fixpacks at the same time
> 2. member_1  uploads the FIXPACK_1 file, then uploads FIXPACK_1’s md5.
> 3. member_2 uploads the FIXPACK_2 md5, then uploads FIXPACK_2’s md5.
> 
> What will happen?
> 
> 3.) Just a curios question 🙏, why we need to send file name to the API?
> 
> Image

Hi @raven-pan, thanks for the review! 🙇‍♂️ 

> 1.) There's a style issue:

Good catch! I should've been more careful. I'll fix it in the upcoming commit. 🙏 

> I think the current verify API design may encounter some edge cases, for example:
> 
> 1. If 2 members upload different fixpacks at the same time
> 2. member_1  uploads the FIXPACK_1 file, then uploads FIXPACK_1’s md5.
> 3. member_2 uploads the FIXPACK_2 md5, then uploads FIXPACK_2’s md5.

We'll need @jdjgya's help for a more precise answer for these questions, but here's my understanding of how the upload process works:

1. The temporary files for fixpacks and firmware are limited to one per data center. If two fixpacks are uploaded at the same time, the second one will always overwrite the first one. The current API implementation wasn't built to handle concurrent users.
2. (and 3) Only one MD5 verification can happen at a time per data center. If another MD5 verification is requested while one is in progress, the backend returns a 409 Conflict error with the following data:
    ```json
    {
      "code": 409,
      "msg": "md5 verification is already in progress",
      "status": "conflict"
    }
    ```
    - By the way, 409 Conflict for the fixpack MD5 verification API happens when there's a duplicate fixpack ID or a concurrent MD5 check. We don't have business error codes to distinguish between them at the moment, so the frontend just shows "A duplicate Fixpack ID was found, or an MD5 checksum is being verified." when it sees 409 Conflict.

> 3.) Just a curios question 🙏, why we need to send file name to the API?

If I remember correctly, it's because the backend can only access the file's content, not its name, so we have to pass the name as a separate parameter.

---

>UI/UX Discussion:
>
>I saw you implemented the browser confirm to remind the user that the file upload isn't complete yet when they manually close the modal, it's great and very user friendly.
>
>Maybe we could also implement this hint when the user tries to close the browser tab, if the fixpack/firmware file is large.

Sure, this makes perfect sense. I'll implement it in the upcoming commit, thanks for the great suggestion! 🚀

steven-chiu-bigstack:
Update:

1. Fix the MD5 verification result table border radius issue:

   image

2. Show a confirmation dialog before leaving page when a file is uploading/uploaded using the [beforeunload event](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event).
   - Too bad the message is not customizable. The current implementation is the best I can do. 😕 

   - https://github.com/user-attachments/assets/4c62a309-8a23-4166-a8cf-f78f1e87bc1a

Issue #399, opened 2025-05-14
[Bug]:  Event charts dropdowns look like being disabled


traviswu-bigstack:
## Describe the bug

As title

## Version

v3.0.0 pre-rc

## To Reproduce

1. Go to 'events'
2. Click on 'chart'
3. See the problem

## Expected behavior

we should make them clear to users

## Screenshots

![Image](https://github.com/user-attachments/assets/6a6dbf9b-e6b1-407e-8927-86e21e18c8c0)

PR #671, opened 2025-09-22
feat: 690 UI supports i18n

### LINKED ISSUE #690: i18n - Project Infra
1. Use Google Sheet as TMS.
2. Set up i18n node script to parse the TMS.
3. Set up i18next.

raven-pan:
Please checkout the ticket first: [Feature - COS UI Supports I18n · Issue \#399 · bigstack-oss/cubecos](https://github.com/bigstack-oss/cubecos/issues/399)

## Which issue(s) this PR fixes

Fixes #690 

---

## Changes

1.) Parse the Google Sheet and transform the rows into `en.json` and `zh-TW.json` by running `pnpm run parse-i18n`.

截圖 2025-09-26 下午3 25 46

2.) Implement the Language Dropdown.

1. The default language is `en`.
2. Users can select `English` or `繁體中文` in the Language Dropdown. 
3. The language preference will be stored in the browser localStorage.

https://github.com/user-attachments/assets/84522599-9f9a-452b-866a-3b0b27b67545

3.) Roughly translated a few pages to test feasibility and gather feedback

1. Home Page
2. Notification Page
3. Sidebar
4. Header
5. Panel Component
6. Percentage Pie Chart Component

## Notices

1.) The i18n features and workflow still under investigation and may change after the kick-off meeting next week.

2.) The `zh-TW` is not well-translated. I only did a rough test to verify feasibility. If the team really cares about this, we can hide the language dropdown.

3.) We should separate ui-library translations like Ant Design, but I haven't implement this pattern yet. This will be handled in the second PR.

Ant Design pattern:

1. [Internationalization](https://ant.design/docs/react/i18n?s=spin)
2. [UNPKG](https://app.unpkg.com/antd@5.27.1/files/locale)

image 15

image 16

4.) The page need to be reloaded when the language changes, since the the web-app still holds the old Dayjs instances.

image

5.) We need to mock localStorage in the future.

Currently, we use `if (typeof window === 'undefined')` to prevent errors in tests as a workaround. We should find a proper way to mock localStorage to avoid test failures.

截圖 2025-09-26 下午3 44 09

6.) README documentation will be added in the next few PRs.

---

### How to Parse I18n from TMS

1. Step 1: Duplicate the `i18nSheetConfig.json.local.example` as `i18nSheetConfig.json.local` and fill in the i18n google sheet ID.
2. Step 2: Put the google credentials file in the project root and rename it to `i18nSheetCredentials.json.local`.
3. Step 3: Run  `pnpm run parse-i18n`. This will parse the `web-app` and `ui-library` i18n worksheets into repo in JSON format. Whenever the translations are updated in the TMS, we may need to rerun this command and rebuild the cube-cos-ui to release the updated translation changes.

Regarding how to set up Google Sheet Integration, checkout this article:
[【Google Sheets API應用】快速產出i18n使用的語系檔小實作](https://pink-learn-frontend.medium.com/google-sheets-api%E6%87%89%E7%94%A8-%E5%BF%AB%E9%80%9F%E7%94%A2%E5%87%BAi18n%E4%BD%BF%E7%94%A8%E7%9A%84%E8%AA%9E%E7%B3%BB%E6%AA%94%E5%B0%8F%E5%AF%A6%E4%BD%9C-b7cdc3ac405a)

#### Support React hooks

In most cases, you just need:

```tsx
const { t } = useTranslation()

return <p>{t('home.overview.chart.title')}</p>
```

Change Language

```tsx
const { i18n } = useTranslation()

return <button onClick={() => i18n.changeLanguage('en')} />          
```

### Support for Plurals

Screenshot 2025-08-25 at 8 39 53 PM

1. Use `{key}_zero` , `{key}_one` and `{key}_other` .
2. We have to drop the [pluralize](https://www.npmjs.com/package/pluralize) package since it calculates plurals dynamically. Some sentences only require certain words to be pluralized, not the entire sentences.

### Support for interpolation

Screenshot 2025-08-25 at 9 09 03 PM

### Support for Component-Level Interpolation

Use the `<Trans />` component for embedding elements such as bold, links or other components (for example: node in a notification message)

image

### Support for Typescript

1. Autocomplete
2. Type Checking

image

Screenshot 2025-08-26 at 12 45 56 AM

Screenshot 2025-08-26 at 12 46 20 AM

Screenshot 2025-08-26 at 12 46 20 AM

### Support Fallback

image 2


packages/cube-frontend-web-app/src/pages/home/utils.ts
```diff
@@ -79,7 +94,7 @@ export const toMetricsChart = (
       .filter((info) => availableRolesSet.has(info.name))
       .map((info) => ({
         ...info,
-        name: upperFirst(info.name),
+        displayName: upperFirst(info.displayName),
```
lizzy-liang-bigstack: Discussion:

Since we already define `displayName` in our dictionary, and `upperFirst` only affects English alphabetic characters (A–Z, a–z), do you think we can remove the `upperFirst` function here? Or are we keeping it mainly for consistency?

packages/cube-frontend-web-app/src/pages/home/overview/_components/NodePanel/NodeTable.tsx
```diff
@@ -105,17 +117,33 @@ export const NodeTable = (props: NodeTableProps) => {
...
+            [NodeStatusEnum.PoweringOn]: 'home.overview.nodes.poweringOn',
+            [NodeStatusEnum.PoweringOff]: 'home.overview.nodes.poweringOff',
+            [NodeStatusEnum.PoweringCycle]: 'home.overview.nodes.poweringCycle',
+            [NodeStatusEnum.Syncing]: 'home.overview.nodes.syncing',
+            [NodeStatusEnum.Unknown]: 'home.overview.nodes.unknown',
+          }
+          const statusDisplay = t(translationMap[status])
+
+          return <CosStatus status={status} statusDisplay={statusDisplay} />
+        }}
```
lizzy-liang-bigstack: Discussion:

Do you think it would make sense to move `translationMap` outside the render function to avoid unnecessary allocations?

Also, perhaps we could provide a fallback for unmapped statuses in case status is ever something unexpected. For example:

```ts
const statusDisplay = t(statusTranslationMap[status] ?? 'home.overview.nodes.unknown');
```

packages/cube-frontend-ui-library/src/components/CosPanel/CosDashboardPanel/CosDashboardPanelHeader.tsx
```diff
@@ -64,7 +61,7 @@ export const CosDashboardPanelHeader = (
         Icon={ChevronRightIcon}
         {...hyperLinkProps}
       >
-        View All
+        {t('component.panel.viewAll', 'View All')}
```
lizzy-liang-bigstack, line 64: Discussion:

I think in most cases we should use resource files so that all text stays centralised. That said, the second parameter of the `t` function could be a fallback in certain situations.

I’m just wondering, in what cases would it make sense to pass in a default value? Do you think we need to draft a small convention for this. What do you think? 🙂

scripts/parseI18n.ts
```diff
@@ -0,0 +1,116 @@
...
+const cosI18nFolderPath = path.resolve(
+  projectRootPath,
+  'packages/cube-frontend-web-app/src/i18n/locales',
+)
+const uiLibraryI18nFolderPath = path.resolve(
+  projectRootPath,
+  'packages/cube-frontend-ui-library/src/i18n/locales',
+)
+
+const supportedLanguage = ['en', 'zh-TW'] as const
```
lizzy-liang-bigstack, line 30: Discussion (minor):

Right now, both `web-app` and `ui-library` share the same `supportedLanguage`, so the script always generates `en.json` and `zh-TW.json` for both. That’s fine for now, but if the packages ever support different languages, maybe we could consider separating them per worksheet. What do you think? 🙂

lizzy-liang-bigstack (approved):
Hi @raven-pan,

Thank you for the amazing work! I’ve learned a lot from this PR 🚀

One question:
I noticed a brief flash after switching languages, just wanted to check if that’s expected. I guess it happens because the page reloads immediately after changing the language. It’s not a big deal, just wanted to confirm.

Feel free to merge once you think it’s ready :)

https://github.com/user-attachments/assets/bee34870-0614-4261-829c-341944c47361

raven-pan:
> r switching languages, just wanted to check if that’s expected. I guess it happens because the page reloads immediately after changing the language. It’s not a big deal, just

Yes, this is expected. The `Dayjs` instance still holds the old locale even after calling `dayjs.locale(language)`.
This causes the UI to display inconsistent languages. For example, if the user switches the language from `zh-TW` to `en`, but the `Dayjs` instance still retains the `zh-TW` locale, then calling the `dayjsInstance.format('HH:mm A')` will return `13:11 上午`, which is unexpected.

The minimal-effort solution is to reload the entire page, but this sacrifices a lot of UI/UX.
I am still looking for a  reasonable solution to handle this.

image

截圖 2025-10-02 下午2 46 37

packages/cube-frontend-web-app/src/pages/home/utils.ts
raven-pan: Yes, I will remove it. Thanks!

packages/cube-frontend-web-app/src/pages/home/overview/_components/NodePanel/NodeTable.tsx
raven-pan: > Do you think it would make sense to move translationMap outside the render function to avoid unnecessary allocations?

Yes!

> Perhaps we could provide a fallback for unmapped statuses in case status is ever something unexpected.

Maybe we can discuss it later. In my opinion, since It already uses `Record<NodeStatusEnum, ParseKeys>` to type-check the `translationMap`, I'd prefer not to affect the business logic with translation. Instead, it should be caught by the `ErrorBoundary`, since it's a totally unexpected condition that requires further investigation by the developers.

packages/cube-frontend-ui-library/src/components/CosPanel/CosDashboardPanel/CosDashboardPanelHeader.tsx
raven-pan: Sorry my bad. This part is not completed yet, and it currently looks very confusing 🙏

In my plan (next PR), it will:
1. Yes, I will centralised translations in the `ui-library/src/i18nlocales/{en,zh-TW}.json`
2. `Not` use any fallback, it should just use `{t('component.panel.viewAll'}`.
3. It should not have any default values. it should only use the keys that already defined in the UI library Google Sheet.
4. If the developer chooses to pass customization locale from `ConfigProvider`, We can enforce that they provide all i18n keys by type checking.

image

Maybe we can discuss this tomorrow. Thanks for your input!

scripts/parseI18n.ts
raven-pan: You are correct. From an architecture aspect, they are separated, and they may support different support languages in the future.

In the ideal solution, it should:
1. The `web-app` and the `ui-library` each have their own worksheet. (Already implemented.)
2. They `can`  support language. (To be achieved  in the future.)
3. The `parse-i18n.ts` script should handle cases where they have different language support. (To be implemented in the next PR.)

PR #716, opened 2025-11-03
feat(integration storage): 588(setDefault) 589(delete) 715(verify) api integration

### LINKED ISSUE #588: [Feature] Storage Integration - Set Default Storage
[Figma](https://www.figma.com/design/kSUzxXJ3OZ5RXSZCRjZ9o1/COS-3.1-M2-UI-mockup?node-id=12113-9732&p=f&m=dev)

Image

Image

Image

Image

Image

### LINKED ISSUE #589: [Feature] Storage Integration - Delete Storage
[Figma](https://www.figma.com/design/kSUzxXJ3OZ5RXSZCRjZ9o1/COS-3.1-M2-UI-mockup?node-id=13707-13310&m=dev)

Image

Notes:

We need a confirmation modal to prevent users from accidentally deleting storage integrations.

### LINKED ISSUE #715: [Feature] Storage Integration - Verify Storage
Image

Image

Image

raven-pan:
# ⚠️ Warning: Do not delete storages. The add/edit storage feature is not yet implemented.

#### What type of PR is this?

feat

feat(integration storage): 588(setDefault) 589(delete) 715(verify) api integration.

Closes #589
Closes #588
Closes #715 

### Screenshots

image

#### Notes

1. To set and external storage as the default, it must be verified first.
2. Built-in storage supports only `setDefault` and does not require prior verification.
4. If a storage has just created/updated, it must be re-verify.
5. The default storage cannot be deleted.

#### UI/UX

1.) The backend handles multiple storage actions via FIFO queue, so executing multiple actions in a short period won't cause an issue. However, users may forget which actions they're already requested and become confused, so I implemented the following action blocks:

1.1) If a `setDefault` action already is processing, all other `set default` button are disabled.

1.2) If any action is processing (create, update, delete, verify, setDefault), then all actions will be blocked in that row.

截圖 2025-11-05 下午4 13 58

2.) To show which action is in progress, I added a spinner with a tooltip

截圖 2025-11-05 下午3 34 28

3.) I added a deletion confirmation modal to prevent accidental storage deleting.

截圖 2025-11-05 下午3 37 25

#### Notifications

Deletion, verify and set default have their notifications.

image

Please see [Toast Notification](https://docs.google.com/spreadsheets/d/196Xy8s6K4PcvpJjQIxX4VAFQzI_ETY2DTbOAODT7AaY/edit?gid=1674446536#gid=1674446536) and [i18n TMS](https://docs.google.com/spreadsheets/d/1fHAjympbsWiAOsgYF84kOD7eKr5p3fREcQiP9I5WtJ0/edit?pli=1&gid=0#gid=0)

#### How to Test Verification?

Since the UI doesn’t support creating or editing storage, we can’t perform a second verification. As a temporary measure, we can modify the code to enable the `Verify` button for testing.

截圖 2025-11-05 下午4 28 17

截圖 2025-11-05 下午4 32 49


packages/cube-frontend-web-app/src/pages/integrations/storages/_components/useStorageTable.tsx
```diff
@@ -0,0 +1,183 @@
...
+        const newSet = new Set(prev)
+        newSet.delete(row.name)
+        return newSet
+      })
+    }
+  }
+
+  const [deleteTargetRow, setDeleteTargetRow] = useState<
+    StorageRow | undefined
+  >(undefined)
```
steven-chiu-bigstack: Nitpick: storing the entire `row` object in state can make the state (UI) of sync with the latest API data. For example:

- When the delete modal opens, `deleteTargetRow` = `{ name: 'dell-emc-sc-fc', isDefault: true }`
- Later, the backend updates this storage to `{ name: 'dell-emc-sc-fc', isDefault: false }`
- In this case, the state will still hold the stale `{ name: 'dell-emc-sc-fc', isDefault: true }`

Since we only need the storage name for deletion, it'd be safer to simplify the state to just `string | undefined` (storage name) instead of the full `StorageRow | undefined`. We can use it alongside with `useMemo` if we really need to access the entire row in the future. But for now, I think we can just remove the `row` property from the `deleteAction` below.

steven-chiu-bigstack (approved):
Hi @raven-pan

Nice work! 🚀 🎉 👍
The current implementation is a bit complex, but it’s functioning correctly and meets requirements. With the deadline approaching, we can leave it as-is and only revisit it if we have to. 🙏

packages/cube-frontend-web-app/src/pages/integrations/storages/_components/useStorageTable.tsx
raven-pan: Sure, your right!
Let me fix it.

PR #717, opened 2025-11-05
feat(integration storage): 585 586 add edit storage api integration

### LINKED ISSUE #585: [Feature] Storage Integration - Add Storage
[Figma](https://www.figma.com/design/kSUzxXJ3OZ5RXSZCRjZ9o1/COS-3.1-M2-UI-mockup?node-id=12113-9732&p=f&m=dev)

Image

Image

### LINKED ISSUE #586: [Feature] Storage Integration - Edit Storage
[Figma](https://www.figma.com/design/kSUzxXJ3OZ5RXSZCRjZ9o1/COS-3.1-M2-UI-mockup?node-id=12113-9732&p=f&m=dev)

Image

Image

raven-pan:
#### What type of PR is this?

feat

feat(integration storage): 585 586 add edit storage api integration

截圖 2025-11-13 下午1 56 24

https://github.com/user-attachments/assets/43986402-5c53-40d5-8cf4-74d59bd96b96

https://github.com/user-attachments/assets/8802a49f-31c9-4bb9-9239-3160776c9b46

feat(integration storage): 585 586 add edit storage api integration

Fixes #585 
Fixes #586 

### Notes

1. I added `immer`  to handle deep-structure field changes and avoid a lot of `…`  in `setState`.
2. The fields (and their default values) in the `driveSection`, `volumeType`, `extraSettings`, `extraConfigFiles`, `imageSettings` are dynamically determined by the select model (called `baseModel` in the code). I call these `ModelBasedSection`.
3. If a field key contains `password`, `secret`, `passwd` or `pwd`, the input field will use `CosPasswordInput` and the value will be displayed as  `***`.
4. When createing storage, if the user changes the `name (driver/model)`, all `ModelBasedSection` are reset to their default values.
5. Only `storageName`, `vendor` and `name(model/driver)` fields are required, and they are disabeld when editing storage. l called these `CommonSection`.
6. All files are base64-encoded before being sent to the backend API.
7. Because extra config files may contains sensitive information and the backend does not return them when fetching existing storage, so users must re-upload these files when editing.

### Known issue

1. ⚠️ Editing storage may cause the whole 45.10 cluster crash.
2. GetStorageByName API has performance issue, it costs about 15 seconds.
3. The order of `driveSection`, `volumeType`, `extraSettings`, `extraConfigFiles`, `imageSettings`  can be discuss with the teams.


packages/cube-frontend-web-app/src/pages/integrations/storages/_components/CreateStorageButton.tsx
```diff
@@ -25,14 +25,14 @@ export const CreateStorageButton = () => {
...
         <CosOverflowMenu.Item
-          key={v.vendor}
+          key={vendor}
           type="plain"
-          title={v.vendor}
+          title={vendor}
           onClick={() =>
             navigate(
-              `${CosRoutesEnum.INTEGRATIONS_STORAGES_CREATE_PAGE}?${DEFAULT_VENDOR_QUERY_KEY}=${v.vendor}`,
+              `${CosRoutesEnum.INTEGRATIONS_STORAGES_CREATE_PAGE}?${DEFAULT_VENDOR_QUERY_KEY}=${vendor}`,
```
steven-chiu-bigstack: Just a note:
Too bad that `react-router`'s `navigate` function does not support passing an object as query params. We could consider using the object format of `to` for complex cases:

image

```ts
import { createSearchParams } from 'react-router'

navigate({
  pathname: CosRoutesEnum.INTEGRATIONS_STORAGES_CREATE_PAGE,
  search: createSearchParams({
    [DEFAULT_VENDOR_QUERY_KEY]: vendor,
  }).toString(),
})
```

packages/cube-frontend-web-app/src/pages/integrations/storages/create/useVendorModel.ts
```diff
@@ -35,10 +60,35 @@ export const useVendorModel = () => {
     }),
   )
 
+  const [selectedVendor, setSelectedVendor] = useState(() =>
+    getVendor(initialStorage, vendorFromQuery),
+  )
+
+  const [selectedModel, setSelectedModel] = useState(() =>
+    getDriver(initialStorage),
+  )
```
steven-chiu-bigstack, line 114: I think we can increase readability by renaming `selectedModel` to `selectedModelName` or `selectedDriver` to distinguish it from the `baseModel` below.

packages/cube-frontend-web-app/src/pages/integrations/storages/create/useVendorModel.ts
```diff
@@ -35,10 +60,35 @@ export const useVendorModel = () => {
...
+  }, [vendorFromQuery, initialStorage])
+
+  useEffect(() => {
+    setSelectedModel(getDriver(initialStorage))
+  }, [initialStorage])
+
   const vendorModels = data ? groupModelsByVendor(data) : {}
+  const vendorOptions = Object.keys(vendorModels)
+  const modelOptions = selectedVendor ? vendorModels[selectedVendor] || [] : []
+  const baseModel = modelOptions?.find((m) => m.driver === selectedModel)
```
steven-chiu-bigstack: - In the current code structure, these array iterating operations are being executed every time storage form gets updated. Even though no performance issue is encountered at the moment considering the amount of vendors and models are low during development, this kind of "looping arrays at every render" still scares me every time. 😨 
- We could rename `baseModel` to `selectedModel` as the intent of "base" is not very intuitive.

packages/cube-frontend-web-app/src/pages/integrations/storages/create/useVendorModel.ts
```diff
@@ -35,10 +60,35 @@ export const useVendorModel = () => {
...
 
   return {
     isLoading,
-    vendorModels,
+    vendorOptions,
+    modelOptions,
+    selectedVendor,
+    selectedModel,
+    setSelectedVendor,
+    setSelectedModel,
```
steven-chiu-bigstack: Nitpick: the code works well, but I think it would be safer in the long run to not expose `setState` functions like `setSelectedVendor` and `setSelectedModel` directly, because:

- We don't have the need to update state based on its current value (i.e. no updater function is needed for `selectedVendor` and `selectedModel`)
- Exposing raw `setState` allows accidental clearing via `setState(undefined)`

But it's just a minor refactor, so it's okay to leave it as is. 🙏

packages/cube-frontend-web-app/src/pages/integrations/storages/_components/upsert/Form/CommonSection.tsx
```diff
@@ -0,0 +1,145 @@
...
+  setSelectedVendor: React.Dispatch<React.SetStateAction<string | undefined>>
+  selectedModel: string | undefined
+  setSelectedModel: React.Dispatch<React.SetStateAction<string | undefined>>
+  vendorOpts: string[]
+  modelOpts: ListIntegrationStorageModelsResponseDataInner[]
+  storage: StorageForm
+  setStorage: React.Dispatch<React.SetStateAction<StorageForm>>
+  formState: StorageFormState
+  fieldErrors: StorageFormError | undefined
+}
```
steven-chiu-bigstack: Note: We can use `Pick` and/or `Omit` with `ReturnType<typeof useVendorModel>` to declare props to save some efforts.

packages/cube-frontend-web-app/src/pages/integrations/storages/_components/upsert/Form/ModelBasedSection/ConfigFileUpload.tsx
```diff
@@ -0,0 +1,52 @@
+import { CosUpload } from '@cube-frontend/ui-library'
+import { fileToBase64 } from '@cube-frontend/web-app/utils/file'
+import { useState } from 'react'
+
+const EXTRA_CONFIG_FILE_SIZE_LIMIT_MB = 10
+const EXTRA_CONFIG_FILE_SIZE_LIMIT = 10 * 1024 * 1024
```
steven-chiu-bigstack: Nitpick: `EXTRA_CONFIG_FILE_SIZE_LIMIT_MB  * 1024 * 1024` would be slightly safer.

packages/cube-frontend-web-app/src/pages/integrations/storages/create/useVendorModel.ts
```diff
@@ -35,10 +60,35 @@ export const useVendorModel = () => {
...
+  )
+
+  useEffect(() => {
+    setSelectedVendor(getVendor(initialStorage, vendorFromQuery))
+  }, [vendorFromQuery, initialStorage])
+
+  useEffect(() => {
+    setSelectedModel(getDriver(initialStorage))
+  }, [initialStorage])
+
```
steven-chiu-bigstack: Do you think it's necessary to update `selectedModel` to the first available option in the vendor-models map when `selectedVendor` changes? To be more specific, when a user explicitly selects a different vendor from the dropdown. 

~~We didn't auto-select the first model during storage creation, so the answer is probably no.~~

On second thought, I think we should, right? This prevents users from seeing stale model data when changing vendors.

packages/cube-frontend-web-app/src/pages/integrations/storages/_components/upsert/Form/ModelBasedSection/ModelBasedSection.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const { baseModel, storage, setStorage, formState, fieldErrors } = props
+  const { service, volumeType } = storage.storage
+  const { driverSection, extraSettings, extraConfigFiles } = service
+  const { settings: volumeTypeSettings } = volumeType
+
+  const showDriverSection = driverSection && driverSection.length > 0
+  const showExtraSettings = extraSettings && extraSettings.length > 0
+  const showExtraConfigFiles = extraConfigFiles && extraConfigFiles.length > 0
+  const showVolumeTypeSettings =
+    volumeTypeSettings && volumeTypeSettings.length > 0
```
steven-chiu-bigstack: Nitpick: It looks like we can remove all the null checks here.

packages/cube-frontend-web-app/src/pages/integrations/storages/_components/upsert/Form/ModelBasedSection/ModelBasedSection.tsx
```diff
@@ -0,0 +1,87 @@
...
+          storage={storage}
+          setStorage={setStorage}
+          formState={formState}
+        />
+      )}
+      {showDriverSection && (
+        <>
+          <CosStroke type="regular" />
+          <DriverSection
+            storage={storage}
```
steven-chiu-bigstack, line 43: Nitpick: For these model-specific sections, I think we only need to pass in the corresponding field as the props rather than the entire `storageForm` object. For example, `<DriverSection>` only needs `driverSection`.
But it's just a minor refactor, so it's okay to leave it as is. 🙏

packages/cube-frontend-web-app/src/pages/integrations/storages/_components/upsert/Form/ModelBasedSection/ExtraSettingsSection.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  const renderSection = (
+    extraSettingsSection: ListIntegrationStorageModelsResponseDataInnerStorageServiceExtraSettingsInner,
+    sectionIndex: number,
+  ) => {
+    const { sectionHeader } = extraSettingsSection
+    return (
+      <div
+        key={sectionHeader}
```
steven-chiu-bigstack: Fields like `sectionHeader` (and `field.key` from key-value pairs) are provided by users, and we don't have a strict duplication check against them, meaning they are not guaranteed to be unique. We can include item index in the (React) `key` prop to avoid unexpected rendering issues if some values do duplicate.

packages/cube-frontend-web-app/src/pages/integrations/storages/_components/upsert/storageFormUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+  if (initialStorage) {
+    return initialStorage.name
+  }
+  return ''
+}
+
+const calculateInitialFields = (
+  initialStorageFields: keyValuePairSchemaType[],
+  baseModelFields: keyValuePairSchemaType[],
+): keyValuePairSchemaType[] => {
```
steven-chiu-bigstack, line 70: This is definitely subjective, but I think renaming `calculateInitialFields` and the arguments would make the intention clearer:

```ts
import { StorageKeyValuePair } from '@cube-frontend/api'

const mergePairsByBaseKeys = (
  basePairs: StorageKeyValuePair[],
  existingPairs: StorageKeyValuePair[],
): StorageKeyValuePair[] => {
  const existingValuesMap = new Map(
    existingPairs.map((pair) => [pair.key, pair.value]),
  )
  return basePairs.map((pair) => ({
    ...basePair,
    key: basePair.key,
    value: existingValuesMap.get(basePair.key) ?? basePair.value,
  }))
}
```

I went with `...basePair` in case additional fields are added to `StorageKeyValuePair` in the future, even though this should not be happening. 🤷‍♂️

packages/cube-frontend-web-app/src/pages/integrations/storages/_components/upsert/storageFormUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  return {
+    useMultipath: false,
+    forceMultipath: false,
+  }
+}
+
+export type GetInitialStorageFormProps = {
+  initialStorage: GetIntegrationStorageResponseData | undefined
+  previousStorage: StorageForm | undefined
```
steven-chiu-bigstack: It looks like `previousStorage` is only here because we want to preserve the `name` field when switching models. If so, I’d lean toward replacing `previousStorage: StorageForm | undefined` with `currentName: string` to keep things lighter.

packages/cube-frontend-web-app/src/pages/integrations/storages/_components/upsert/storageFormUtils.ts
```diff
@@ -0,0 +1,230 @@
...
+  return baseModelFields.map((modelField) => {
+    const existingField = initialStorageFieldsMap.get(modelField.key)
+    return existingField ?? modelField
+  })
+}
+
+const getInitialDriverSection = (
+  initialStorage: GetIntegrationStorageResponseData | undefined,
+  baseModel: ListIntegrationStorageModelsResponseDataInner | undefined,
+) => {
```
steven-chiu-bigstack, line 85: To keep things clearer, I’d suggest passing down only the values we need, instead of the entire model and API response. For example:

```ts
const getInitialDriverSection  = (
  baseDriverSection: ListIntegrationStorageModelsResponseDataInnerMultipathInnerAttributesInner[],
  existingDriverSection: StorageKeyValuePair[],
): StorageKeyValuePair [] => {
  return something // Could probably reuse the `mergePairsByBaseKeys` mentioned above.
}

const driverSection = getInitialDriverSection(
  baseModel?.storage.service.driverSection ?? [],
  initialStorage?.storage.service.driverSection ?? [],
)
```

If we want to be extra safe, we should probably treat the existing values (i.e. values from `initialStorage`) as `unknown`, because who knows how the schema will change in the future. 🤷‍♂️

Updated fields could end up looking completely different from the old ones. For example, if a field that used to be an array becomes an object, we wouldn't be able to catch that in static checks and would blow up at runtime with something like “arg.map is not a function”.

That said, I don’t think that's necessary at the moment. In such scenario, there should be a migration strategy in the backend.

steven-chiu-bigstack (approved):
Hi @raven-pan 

Looks good! As always, really appreciate your great work! 🚀 🙏 🙇‍♂️

packages/cube-frontend-web-app/src/pages/integrations/storages/create/useVendorModel.ts
raven-pan: Sure!

packages/cube-frontend-web-app/src/pages/integrations/storages/_components/CreateStorageButton.tsx
raven-pan: Sure, that's much better.
Let me fix it.

packages/cube-frontend-web-app/src/pages/integrations/storages/create/useVendorModel.ts
raven-pan: Sure!

packages/cube-frontend-web-app/src/pages/integrations/storages/create/useVendorModel.ts
raven-pan: 1. Sure, let me optimize it. 
2. Sure.

packages/cube-frontend-web-app/src/pages/integrations/storages/create/useVendorModel.ts
raven-pan: Your right, this is a good habit. Let me refactor it.

packages/cube-frontend-web-app/src/pages/integrations/storages/_components/upsert/Form/ModelBasedSection/ExtraSettingsSection.tsx
raven-pan: Thanks for reminding me.

Under the normal flow, duplicated aren't expected; however, we don't have a strict duplication check. Using the item index is much safer.

raven-pan:
Hi @steven-chiu-bigstack 

Thanks for your review. I've fixed most comments and added the confirmation modals for the edit storage, replace model and replace model list actions. Design has added these in Figma as well.

截圖 2025-11-18 晚上7 01 11

截圖 2025-11-18 晚上7 01 34

steven-chiu-bigstack:
Got it, thanks! Feel free to merge the PR when your're ready. 🙏

PR #749, opened 2025-12-16
748 notifications page check link url update

### LINKED ISSUE #748: [Enhancement] Notifications Page "Check" Link URL Update
Currently the "Check" link for all notifications navigate users to Node Details page. This don't make sense anymore as there are some other types of notifications being added (e.g., storage, model, etc.)

Compute the most suitable link URL based on notification id.

None

**Additional context**
Slack discussion: https://bigstack-tech.slack.com/archives/C07GL6RKCHM/p1765779854859679

steven-chiu-bigstack:
#### What type of PR is this?

Feat

Compute notifications link URL based on notification id.

Fixes #748

The solution may seem over-complicated compared to `notification.id.startsWith('...')`, but I think it'd be better to make the code more sensitive to API type changes so that static checks (`tsc`) could catch problems as early as possible (e.g., when `MDL00001E` is changed to `MOD00001E` in the future.)

#### Additional documentation

https://github.com/user-attachments/assets/553c2a12-12c0-4928-b0d0-f0899279f42d


raven-pan (approved):
LGTM!

I think your solution is perfect and totally worth it, which leverages Typescript to protect our codebase if there are new changes in the OpenAPI.

PR #764, opened 2026-01-08
feat(i18n): sync google sheet using ADC

raven-pan:
feat(i18n): sync google sheet using ADC 

#### What type of PR is this?

feat, doc

1. feat(i18n): sync google sheet using ADC
2. docs(i18n): workflow and guidelines.
3. feat(i18n): rename all `執行個體` to `虛擬機`


packages/cube-frontend-i18n/docs/README.md
```diff
@@ -0,0 +1,212 @@
+# CMP i18n Translation Workflow
```
steven-chiu-bigstack: It looks like "CMP" should be renamed to "COS".

packages/cube-frontend-i18n/docs/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+- Script location:
+
+  ```bash
+  scripts/syncI18n.ts
+  ```
+
+- How to run:
+
+  From the i18n director, execute:
```
steven-chiu-bigstack: Did you mean to say "directory" instead of "director"?

packages/cube-frontend-i18n/docs/README.md
```diff
@@ -0,0 +1,212 @@
...
+## I18n Key Format
+
+### Web app
+
+1. `page + place/context (optional) + label/function`
+2. `layout (sidebar/header) + place/context (optional) + label/function`
+
+Examples:
+
+`overview.home.chart.vmSummary`
```
steven-chiu-bigstack: Just curious, the "Use Camel Case" sections below uses `home.overview.chart.vmSummary` as an example, but here it's `overview.home.chart.vmSummary`. Do you think it's necessary to unify them?

packages/cube-frontend-i18n/docs/README.md
```diff
@@ -0,0 +1,212 @@
...
+1. We have to define another i18n key with `_{others}` suffix.
+2. We must defined the name of the placeholder as `count`
+
+| Key                                  | en                          | zh-TW                     |
+| ------------------------------------ | --------------------------- | ------------------------- |
+| home.overview.instanceCount          | We have {{count} instance.  | 我們有 {{count}} 個虛擬機 |
+| home.overview.instanceCount_{others} | We have {{count}} instances | 我們有 {{count}} 個虛擬機 |
+
+1. `home.overview.instance` => singular
+2. `home.overview.instance{other}` => plural
```
steven-chiu-bigstack: Should we update these keys to make them match with the ones in the table above?

packages/cube-frontend-i18n/docs/README.md
```diff
@@ -0,0 +1,212 @@
...
+Reference: [i18next - Trans Component](https://react.i18next.com/latest/trans-component)
+
+### Embedded Link
+
+1. Use `<link>target text</link>` to add link to the target text.
+2. We can embed multiples link in 1 sentence by individual `link1, link2…`.
+
+Examples:
+
+1. `About our license, please check out <link>documentation</link>`.
```
steven-chiu-bigstack: Unfortunately, `link` is one of the [self-closing tags](https://developer.mozilla.org/en-US/docs/Glossary/Void_element) that won't work in `Trans` ([docs](https://react.i18next.com/latest/trans-component#alternative-usage-which-lists-the-components-v11.6.0)). I've run into this problem in CMP, and it took me a while to realize what was happening. 🫤 

{2EB132FA-4F8A-4364-88AD-64B0E5D64366}

packages/cube-frontend-i18n/scripts/syncI18n.ts
```diff
@@ -85,32 +80,30 @@ const parseWorksheet = async ({
...
+  const client = await auth.getClient()
+
+  const doc = new GoogleSpreadsheet(sheetId, client)
   await doc.loadInfo()
 
-  const parseWorksheetTasks = [
+  const syncWorksheetTasks = [
     {
-      workSheet: doc.sheetsByTitle['COS 3.1'],
+      workSheet: doc.sheetsByTitle[version],
```
steven-chiu-bigstack: Is it correct that this should be `` doc.sheetsByTitle[`${version} web-app`] ``?

packages/cube-frontend-i18n/scripts/syncI18n.ts
```diff
@@ -85,32 +80,30 @@ const parseWorksheet = async ({
...
       folderPath: cosI18nFolderPath,
     },
     {
-      workSheet: doc.sheetsByTitle['COS 3.1 UI Library'],
+      workSheet: doc.sheetsByTitle[`${version} UI Library`],
```
steven-chiu-bigstack: The title in the Google Sheet is in kebab-case, so I think `UI Library` should also be changed to kebab-case:

{DF09CB8B-8D56-43D2-A36E-4D3DAEB6200D}

packages/cube-frontend-i18n/package.json
```diff
@@ -3,7 +3,8 @@
   "private": true,
   "license": "Apache-2.0",
   "scripts": {
-    "tsc": "tsc -b"
+    "tsc": "tsc -b",
+    "sync": "node --experimental-strip-types scripts/sync.ts"
```
steven-chiu-bigstack: The file name is actually `syncI18n.ts` instead of `sync.ts`

steven-chiu-bigstack (approved):
Hi @raven-pan 

Nice work, thanks for taking care of this! 🚀

packages/cube-frontend-i18n/scripts/syncI18n.ts
raven-pan: Fixed.

packages/cube-frontend-i18n/scripts/syncI18n.ts
raven-pan: fixed.

packages/cube-frontend-i18n/docs/README.md
raven-pan: yes, fixed.

packages/cube-frontend-i18n/docs/README.md
raven-pan: Sure

packages/cube-frontend-i18n/docs/README.md
raven-pan: @steven-chiu-bigstack 

I use hyperlink, hyperlink1, and hyperlink2 to prevent this pitfall. I also added a warning to our i18n doc like this:

image

packages/cube-frontend-i18n/package.json
raven-pan: Fixed.

PR #832, opened 2026-06-30
refactor: refine gpu cards api response

raven-pan:
#### What type of PR is this?

refactor

N/A

OpenaAPI Changes: https://github.com/bigstack-oss/cube-cos-openapi/pull/104

#### Additional documentation

截圖 2026-06-30 晚上7 17 34


packages/cube-frontend-web-app/src/mocks/handlers.ts
```diff
@@ -1,26 +1,26 @@
...
+  //       msg: 'Success',
+  //       status: 'success',
+  //     }
 
-      return HttpResponse.json(res)
-    },
-  ),
+  //     return HttpResponse.json(res)
+  //   },
+  // ),
```
steven-chiu-bigstack: Nitpick: A lot of code was commented out instead of being removed from this file. I think it would scale better if we moved that code into separate files. For example:

```ts
import { mockListNodeGpuCards, mockUpdateNodeGpuCard  } from './gpu'

export const handlers = [
  mockListNodeGpuCards,
  mockUpdateNodeGpuCard,
]
```

This way, if we want to enable/disable mocking `listNodeGpuCards`, we only have to change a few lines (importing/removing the mock functions).

steven-chiu-bigstack (approved):
LGTM!

steven-chiu-bigstack (approved):

packages/cube-frontend-web-app/src/mocks/handlers.ts
raven-pan: Sure, let me refactor it.

PR #838, opened 2026-07-20
fix: mint GPU instance console link on demand

raven-pan:
#### What type of PR is this?

/kind feature

GPU-attached instance console links are no longer embedded in the `listNodeGPUCards` response. They are now minted on demand per instance via the new `getGpuInstanceConsole` endpoint, opened in a new tab on click; any failure surfaces as an error toast. List `links` now carry only the Grafana deep-link. This avoids minting a Nova console token per instance on every listing poll.

- Bump `cube-cos-openapi` submodule to `develop` and regenerate the SDK (console endpoint, `degraded` flag, `links` without `console`, profile-id type fix).
- New `GpuConsoleLink` component: click → `getGpuInstanceConsole` → `window.open`; disabled while loading; error toast. Dedupes the two identical console render spots.
- `GpuDetails` / `FullViewInstanceTable`: swap the console hyperlink for `GpuConsoleLink`; Grafana unchanged. `nodeName` threaded down from `NodeGpuResources`.
- `editGPUResourceUtils`: update-payload profile `id` now `Number(...)` to match the SDK type flip (string → number).
- Mocks: drop `console` from mock links, add `degraded: false` to cards, add + register `mockGetGpuInstanceConsole`.

Out of scope: rendering the new `degraded` card flag in the UI.

Fixes #

> The submodule points at `cube-cos-openapi` **develop**. Land that spec change before merging so the SDK diff matches a pinned ref.

Verification: `pnpm tsc` clean, `pnpm web-app:test` 98 passed, `pnpm lint` clean.

#### Additional documentation

```docs
Handbook: kb/cubecos/architecture/gpu-card-apis.md — list / update / console endpoint contract.
```

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/GpuConsoleLink.tsx
+++ b/packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/GpuConsoleLink.tsx
@@ -1,0 +1,54 @@
+import { useContext } from 'react'
+import { useTranslation } from 'react-i18next'
+import { CosHyperlink } from '@cube-frontend/ui-library'
+import { DataCenterContext } from '@cube-frontend/web-app/context/DataCenterContext'
+import { nodesApi } from '@cube-frontend/web-app/api/cosApi'
+import { useCosMutationRequest } from '@cube-frontend/web-app/hooks/useCosRequest/useCosMutationRequest'
+import { useShowErrorToast } from '@cube-frontend/web-app/hooks/useShowErrorToast/useShowErrorToast'
+
+export type GpuConsoleLinkProps = {
+  nodeName: string
+  instanceId: string
+}
+
+/**
+ * Console links are minted on demand rather than embedded in the GPU cards
+ * listing (so polling the listing does not create a Nova console token per
+ * instance on every request). Fetch the link on click and open it in a new tab.
+ */
+export const GpuConsoleLink = (props: GpuConsoleLinkProps) => {
+  const { nodeName, instanceId } = props
+
+  const { t } = useTranslation()
+
+  const { dataCenter } = useContext(DataCenterContext)
+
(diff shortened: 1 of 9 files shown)


packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/GpuConsoleLink.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const { nodeName, instanceId } = props
+
+  const { t } = useTranslation()
+
+  const { dataCenter } = useContext(DataCenterContext)
+
+  const showErrorToast = useShowErrorToast()
+
+  const { isLoading, mutateResource: getGpuInstanceConsole } =
+    useCosMutationRequest(nodesApi.getGpuInstanceConsole)
```
Comment (steven-chiu-bigstack) [inline packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/GpuConsoleLink.tsx]: I'd suggest using `useCosGetRequest(nodesApi.getGpuInstanceConsole, { fetchOnMount: false, fetchOnParamChanges: false })` for "getting" console link. The API data can be retrieved via `getResource()` instead of relying on render-determined `data`.

steven-chiu-bigstack (approved):
LGTM!

packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/GpuConsoleLink.tsx
Comment (raven-pan) [reply packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/GpuConsoleLink.tsx]: Switched to `useCosGetRequest(nodesApi.getGpuInstanceConsole, getParam, { fetchOnMount: false, fetchOnParamChanges: false })` and fetch via `getResource()` on click (c49e13e6). One note: the GET handler starts `isLoading: true` and never resets while auto-fetch is off, so the in-flight disabled state uses a local `isOpeningConsole` flag rather than the hook's `isLoading`.

PR #847, opened 2026-08-12
feat(login): upgrade keycloak from 17.0.1-legacy to 18.0.2 quarkus

Eandalf-Bigstack:
#### What type of PR is this?

/kind feature

Moves the `cube-cos-login` image off Keycloak `17.0.1-legacy` — the last of the WildFly
line, dating from 2022 — onto `18.0.2`, which runs on Quarkus.

This is the `cube-cos-ui` half of the upgrade. It changes what the login image *is*; the
`cubecos` side changes how it is deployed and configured.

**The image (`KEYCLOAK_VERSION`, `cube-cos-login.spec`, `package.json`)**

- Base image `quay.io/keycloak/keycloak:17.0.1-legacy` → `18.0.2`. Note that an
  `18.0.2-legacy` tag also exists and would have kept us on WildFly; the point of the story
  is the Quarkus move, so we take the plain tag.
- Themes are copied to `/opt/keycloak/themes/` instead of `/opt/jboss/keycloak/themes/`.
- The spec now runs `kc.sh build` at image build time for `db`, `cache`, `health-enabled`,
  `metrics-enabled` and `http-relative-path`. Quarkus fixes these when the server is
  augmented, so leaving them to the runtime environment makes **every pod re-augment while
  starting**, and `http-relative-path` cannot be changed at runtime at all — a pod given it
  only as an env var comes up serving `/` instead of `/auth`. Pre-baking also drops pod
  startup to under 7s.
- `cache-ispn.xml` ships with two cache owners, so the spec rewrites it to three, preserving
  the `CACHE_OWNERS_COUNT=3` the WildFly deployment asked for.
- Package version `17.2.0` → `18.0.0`. This value is both the RPM version and the image tag.

**The local preview (`docker-compose.yaml`, `Dockerfile`, `keycloak/standalone/`)**

The `Dockerfile` existed only to drop `standalone.xml` / `standalone-ha.xml` over the WildFly
ones, whose sole purpose was disabling the theme cache. Quarkus has no `standalone.xml`, and
`start-dev` already turns theme caching off, so all three files and the image build they
needed are gone — `docker-compose` runs the upstream image directly.

Environment renames follow the distribution: `KEYCLOAK_USER` / `KEYCLOAK_PASSWORD` →
`KEYCLOAK_ADMIN` / `KEYCLOAK_ADMIN_PASSWORD`, theme mount path change, and
`KC_HTTP_RELATIVE_PATH=/auth` so local URLs match how CubeCOS serves them.

Part of bigstack-oss/cubecos#621

Not a `Fixes`, because the story is only complete once the `cubecos` side lands as well.

> **The theme itself needed no porting.** `login.ftl`, `template.ftl` and `theme.properties`
> render unchanged on 18.0.2. `parent=base`, `import=common/keycloak`, and every variable the
> templates use (`url.resourcesPath`, `url.loginAction`, `messagesPerField.existsError`,
> `auth.selectedCredential`, `realm.rememberMe`, `usernameHidden`) all still resolve. The
> `${env.LOGIN_GREETING:}` substitution in `theme.properties` also still works under Quarkus,
> including the empty default — verified it renders as `loginGreeting: ''` when unset and
> carries the value through when set.

> **`KC_SPI_THEME_DEFAULT` in `docker-compose.yaml` is a dev convenience, not what production
> does.** It is a server-wide default covering every theme type, and this package only ships
> a `login` theme, so a preview container logs `Failed to find WELCOME theme cos-ui`. That is
> harmless locally and it means a developer sees the COS login page without first changing a
> realm setting. CubeCOS deliberately does **not** deploy it that way — it sets the theme as
> the master realm's login theme, which is what bigstack-oss/cubecos#187 asked for. The
> README spells out the difference and how to mirror production locally.

> **`buildah run` is new in the spec** (for the `sed` and the `kc.sh build`). Confirmed it
> works in the build jail: the login RPM is built as root there, where `buildah run` has a
> working OCI runtime. It is not affected by the rootless user-namespace limitation that
> would bite if the build ever moved to the `jenkins` user.

> **Verification.** The React app was not rebuilt for this — nothing under `src/` changed — so
> the image was validated by rebuilding it with the real theme payload (882 files including
> the built css/js/img/fonts) taken out of the currently shipping `17.2.0` image. That image
> then migrated a live Keycloak 17 schema to `18.0.0-10625` on the `jim-1cc` e2e node, and the
> `cos-ui` login page rendered with `css/index.css`, `js/index.js`, `img/favicon.svg` and the
> Inter woff2 fonts all answering 200. A full PKCE authorization-code login through the admin
> console also completed against it.

#### Additional documentation

```docs
packages/cube-frontend-keycloak-login/README.md
```


steven-chiu-bigstack (approved):
LGTM! 🚀
I've tested it manually, and everything works as expected.
It looks like the theme caching issue does not exist in version 18. The entire `keycloak/standalone/configuration` folder is gone.

PR #850, opened 2026-08-17
chore(workflow): update actions workflow with hardened runner

bigstack-brian-su:
#### What type of PR is this?
 
Chore

Update workflow permissions and harden runner.


steven-chiu-bigstack (approved):
LGTM!

PR #823, opened 2026-05-22
feat: 869 gpu listing page full view modal

lizzy-liang-bigstack:
#### What type of PR is this?

feat

#### What this PR does / why we need it

1. Add `full_screen` icon to UI library

Screenshot 2026-05-22 at 1 47 16 PM

2. Add `Full View Modal` to GPU resource listing page

https://github.com/user-attachments/assets/826024ed-04ec-42f9-8a91-f64d2f408a1c

#### Which issue(s) this PR fixes

Fixes https://github.com/bigstack-oss/cubecos/issues/868

#### Special notes for your reviewer

> Note

#### Additional documentation

```docs

```

PR #852, opened 2026-08-17
feat(nodes): 824 open the Grafana GPU history panels from a GPU row

raven-pan:
#### What type of PR is this?

/kind feature

An operator can now jump from a physical GPU card to that card's history in
Grafana. The row overflow menu in **Nodes > \<node\> > GPU Resources** carries two
new entries:

| Menu entry | Opens | Ticket |
| --- | --- | --- |
| View Workload History | GPU Utilization over time, for that card | #824 |
| View VRAM History | VRAM usage over time, for that card | #825 |

The links arrive **inline with the GPU card listing** — `links.workloadHistory`
and `links.vramHistory` on every card — so a row action costs no extra request,
and each URL already carries the card's PCI address (`var-GPU_PCIID`). The UI
opens `link` verbatim in a new tab; it never builds a URL, so panel ids stay
inside `cube-cos-api`.

An entry is disabled when the card's resource type has no series to draw:

| Resource type | Workload | VRAM | Why |
| --- | --- | --- | --- |
| unset, sriovVgpu | enabled | enabled | the host reads both |
| migBackedVgpu | disabled | enabled | NVIDIA reports no device-level utilization once MIG is on |
| pgpu | disabled | disabled | the card is bound to vfio-pci, so the host sees nothing |

The API returns a link for those cards too — correctly, since the API should not
decide UX. The UI still greys the entry out, because that chart can never fill,
and the same row already shows `-` for the matching live stat.

This is the last track of #824 and #825. Infra (panels + hidden `$GPU_HOST` /
`$GPU_PCIID`), the collector, and the backend are handled in their own repos.

##### Screenshots

All three shots come from one node with the MSW GPU mocks, which cover every
resource type.

**1. Unset row — both history entries enabled**

1-unset-both-enabled

**2. MIG-backed vGPU row, in use — Workload History greyed out, VRAM History still live**

2-mig-backed-vgpu-workload-disabled

**3. Passthrough row, idle — both history entries greyed out, Edit GPU Type still live**

3-passthrough-both-disabled

Fixes #824

##### Related PRs

Every dependency is merged. The UI only reads what these produced.

| Repo | PR | What it gives this PR |
| --- | --- | --- |
| cube-cos-openapi | [#108](https://github.com/bigstack-oss/cube-cos-openapi/pull/108) | `links.workloadHistory` / `links.vramHistory` on each GPU card |
| cube-cos-openapi | [#107](https://github.com/bigstack-oss/cube-cos-openapi/pull/107) | the nullable GPU stats that come with the same submodule bump |
| cube-cos-openapi | [#109](https://github.com/bigstack-oss/cube-cos-openapi/pull/109) | a nullable Grafana link on an attached instance |
| cube-cos-api | [#636](https://github.com/bigstack-oss/cube-cos-api/pull/636) | builds both per-card links, filtered by `var-GPU_PCIID` |
| cube-cos-api | [#635](https://github.com/bigstack-oss/cube-cos-api/pull/635) | reports unmeasurable stats as `null` instead of `0` |
| cubecos | [#1019](https://github.com/bigstack-oss/cubecos/pull/1019) | panels `50` / `51` and the hidden `$GPU_HOST` in `device.json` |
| cubecos | [#1228](https://github.com/bigstack-oss/cubecos/pull/1228) | the collector fix that lets those panels hold data at all |

Follow-up in this repo: [#853](https://github.com/bigstack-oss/cube-cos-ui/pull/853) turns the blank cells this PR leaves behind into a dash plus a reason (#823).

> Note

- **Both dependencies are merged.** The submodule tracks `cube-cos-openapi`
  develop (`c865681`), which carries bigstack-oss/cube-cos-openapi#108;
  bigstack-oss/cube-cos-api#636 implements the endpoint side. Nothing here waits
  on another repo.
- **develop also carries the #823 spec change**, which marks the unmeasurable GPU
  stats and an instance's Grafana link nullable. You cannot take the per-card
  links without it — that commit landed first. The last commit is the null-safety
  this forces: a value the host cannot measure renders as an empty cell (today's
  behaviour for a card with no `vram` object), and the instance Grafana link
  renders only when the API supplies one. **Telling the operator why a number is
  missing is still #823's story, not this PR's.**
- **The first two commits used device-scoped endpoints** (`.../grafana/devices/{hostname}/gpuUtilization`
  and `.../gpuVram`), which returned one link per node — so every row of a node
  opened the same chart. Later commits replace that with the per-card links. The
  device-scoped endpoints stay in the spec but have no consumer here.
- Both MSW mocks for the old endpoints are deleted; `src/mocks/gpu.ts` now builds
  each card's links from its PCI address and stays unregistered, matching the
  other GPU mocks.
- The two i18n keys live in the JSON resources only. Add the rows to the
  translation sheet, or the next `pnpm i18n:sync` drops them.

Verified with the MSW mocks on a node carrying 12 cards across all four resource
types: all 12 enabled entries open `var-GPU_PCIID=<that row's own PCI address>`
with `viewPanel=50` / `51` in a new tab (`noopener,noreferrer`), no two rows share
a URL, every disabled entry opens nothing, and the console stays clean. `pnpm
lint` passes and `pnpm test` reports 106 passing tests, 5 of them new.

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-api/sdk/api.ts
+++ b/packages/cube-frontend-api/sdk/api.ts
@@ -3895,6 +3895,56 @@ export interface GetGrafanaDashboardLinkResponseData {
 /**
  * 
  * @export
+ * @interface GetGrafanaDeviceGpuUtilization500Response
+ */
+export interface GetGrafanaDeviceGpuUtilization500Response {
+    /**
+     * 
+     * @type {number}
+     * @memberof GetGrafanaDeviceGpuUtilization500Response
+     */
+    'code'?: number;
+    /**
+     * 
+     * @type {string}
+     * @memberof GetGrafanaDeviceGpuUtilization500Response
+     */
+    'msg'?: string;
+    /**
+     * 
+     * @type {string}
+     * @memberof GetGrafanaDeviceGpuUtilization500Response
+     */
+    'status'?: string;
+}
+/**
+ * 
+ * @export
+ * @interface GetGrafanaDeviceGpuVram500Response
+ */
+export interface GetGrafanaDeviceGpuVram500Response {
+    /**
+     * 
+     * @type {number}
+     * @memberof GetGrafanaDeviceGpuVram500Response
+     */
+    'code'?: number;
+    /**
+     * 
+     * @type {string}
+     * @memberof GetGrafanaDeviceGpuVram500Response
+     */
+    'msg'?: string;
+    /**
+     * 
+     * @type {string}
+     * @memberof GetGrafanaDeviceGpuVram500Response
+     */
+    'status'?: string;
+}
+/**
+ * 
+ * @export
  * @interface GetGrafanaHosts500Response
  */
 export interface GetGrafanaHosts500Response {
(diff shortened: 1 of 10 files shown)


steven-chiu-bigstack (approved):
LGTM!

PR #855, opened 2026-08-19
feat(advisor-ui): the Advisor UI package, and the authority chain

### LINKED ISSUE #854: [Task] Advisor UI package: the shell, starting with the authority chain
## Description

Stand up `packages/cube-advisor-ui` — the Cube AI Advisor SaaS web UI — inside this monorepo, and implement the first cross-cutting element from the redraw brief: the **authority chain**.

The Advisor's UI is drawn from the same design language as CubeCOS, and the [UI brief](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cube-ai-advisor/blueprints/cube-ai-advisor-ui-brief.md) states the same core is later embedded into the CubeCOS product for customers. Living beside `@cube-frontend/ui-library` is what makes that embedding cheap.

## Requirements

- [ ] `packages/cube-advisor-ui` scaffolded to the conventions of the existing packages (pnpm workspace, catalog deps, `tsc`, vitest)
- [ ] `AuthorityChain` — the persistent banner naming every party acting inside a tenant
- [ ] A three-party chain must never render, or summarise, as a two-party relationship
- [ ] The full chain is in the accessibility tree regardless of viewport
- [ ] Colours come from `cubeTheme.ts`, not hand-picked hexes
- [ ] The four "keep extraction cheap" rules hold, with rule 1 enforced by lint

### Keeping the exit door open

The Advisor may later need its own repository with a dedicated frontend team. Extraction stays a few days of work rather than a rewrite only if four things stay true, so they are rules in the package README rather than intentions:

1. Import through package names, never sibling source paths — **enforced** by extending the existing `no-restricted-imports` pattern to cover `cube-advisor*` (it previously only matched `cube-frontend*`, so the new package would not have been caught).
2. Every dependency declared in this package, never borrowed from the workspace root.
3. The backend boundary is an OpenAPI artifact, not shared types.
4. The dependency edge points one way: `advisor-ui → ui-library`.

Worth recording so it is not a surprise later: `@cube-frontend/ui-library` and `@cube-frontend/ui-theme` are `private`, unversioned, and export raw TypeScript (`main: ./src/index.ts`). A second repository cannot consume them as source, so extraction would require giving them a build, a version and an `exports` map — work owed the moment *any* second repo wants those components, not created by this package.

### Why the authority chain first

The redraw calls it "the single most important new element", and it is the one with a hard, testable constraint: a three-party chain *may* truncate visually but must **never** truncate to something that reads as a two-party relationship. `Acme ← NorthWind` reads as direct support by a company called NorthWind — the exact misreading the element exists to prevent.

So the narrow-viewport path is not a truncation of the party list. Dropping the middle party is the naive implementation and is precisely the bug; instead the chain collapses to a hop count that still names the partner tier.

## Output artifacts (Definition of Done)

- [ ] **Handbook knowledge update** — land the durable, team-readable knowledge from this
      work into the bigstack-handbook cube-ai-advisor kb (`kb/cube-ai-advisor/…`).

traviswu-bigstack:

Stands up `packages/cube-advisor-ui` — the Cube AI Advisor SaaS web UI — and implements the first cross-cutting element from the [redraw brief](https://github.com/bigstack-oss/bigstack-handbook/blob/develop/kb/cube-ai-advisor/blueprints/): the **authority chain**.

Fixes #854

## Why the UI lives in this monorepo

The brief says the same UI core is later embedded into the CubeCOS product for customers. Living beside `@cube-frontend/ui-library` is what makes that embedding cheap.

We discussed whether that traps us if the Advisor later needs a dedicated frontend team and its own repo. It doesn't — but only because of a property this repo already has, and which this PR keeps true.

**The evidence:** across the existing web app there are 564 cross-package imports and **every one goes through a package name** (`@cube-frontend/ui-library`, `/api`, `/utils`, `/ui-theme`). Zero deep source imports. Deep imports are what actually weld a package into a monorepo; that discipline is already in place, and `workspace:*` → a version range is a one-line change per dependency at extraction time.

So four rules live in the package README, and **rule 1 is enforced rather than remembered**:

| Rule | How it holds |
|---|---|
| Import through package names, never sibling paths | `no-restricted-imports` — **extended in this PR**, since the pattern only matched `cube-frontend*` and would not have caught the new package |
| Every dependency declared here | pnpm doesn't flat-hoist, so an undeclared dep breaks loudly |
| Backend boundary is an OpenAPI artifact, not shared types | convention, same as `cube-cos-api` ↔ `cube-cos-openapi` |
| Dependency edge points one way: `advisor-ui → ui-library` | review |

**Recorded so it isn't a surprise later:** `@cube-frontend/ui-library` and `@cube-frontend/ui-theme` are `private`, carry no `version`, and `main` points at raw `./src/index.ts`. A second repo can't consume them as source, so extraction would mean giving them a build, a version and an `exports` map. That's work owed the moment *any* second repo wants those components — deferred by this choice, not created by it.

### The authority chain's one hard constraint

The redraw calls it "the single most important new element". Its constraint is unusually testable:

> it may truncate visually but must never truncate to something that *reads* as a two-party relationship

`Acme Corp ← NorthWind` reads as direct support by a company called NorthWind. That misreading is the whole reason the element exists.

So **the narrow-viewport path is not a truncation of the party list.** Dropping the middle party is the naive implementation and is precisely the bug. Instead the chain collapses to a hop count that still names the tier:

```
wide:   Acme Corp ← Partner: NorthWind ← Bigstack L3 · expires in 5h 12m
narrow: Acme Corp ← via 1 partner ← Bigstack L3
```

Two supporting decisions: the full chain is in the accessibility tree whatever the viewport renders, and there is no viewer for whom a party is omitted — a chain that renders differently per audience is not an audit trail.

### Mutation testing

Four mutations, each caught:

- **the naive middle-drop** — three parties rendered as two. This is the bug the brief warns about, so it is the one the test exists for.
- the `Partner:` label removed
- the hop count replaced with a vague word
- the chain order reversed, which would name the wrong party as the data owner

### A lint catch worth mentioning

My first pass invented theme token names (`text-functional-text-01`, `text-status-warning`). `tailwindcss/no-custom-classname` rejected them and I replaced them with the real tokens from `cubeTheme.ts` (`functional-title`, `functional-text-light`, `status-paused`). The rule did exactly its job.

## Verification

`tsc` clean · `eslint` clean (0 problems) · 9 tests pass · prettier applied.

## Not in this PR

The remaining shell elements — **scope switcher / breadcrumb** and the **provider badge** with its new mode dimension — and all twelve screens. The design bundle in the Claude Design project is a *Figma export* (128 components named from layer names), not this library; the screens have to be translated onto the real `Cos*` vocabulary rather than ported, or the design system forks.

### DIFF
--- a/packages/cube-advisor-ui/src/components/AuthorityChain/AuthorityChain.tsx
+++ b/packages/cube-advisor-ui/src/components/AuthorityChain/AuthorityChain.tsx
@@ -1,0 +1,92 @@
+import classNames from 'classnames'
+
+import {
+  formatParty,
+  summariseChain,
+  type AuthorityChainProps,
+  type AuthorityParty,
+} from './formatChain'
+
+/**
+ * The persistent banner shown while anyone is acting inside a tenant they do
+ * not own.
+ *
+ * Whoever is acting reads the same chain the data owner reads. There is no
+ * viewer for whom a party is omitted — an authority chain that renders
+ * differently per audience is not an audit trail.
+ */
+/**
+ * Colours come from cubeTheme.ts, not from this file. A partner is the tier a
+ * reader most needs to notice, so it takes the one accent here; the other two
+ * stay in the ordinary text ramp so the chain reads as one sentence.
+ */
+const tierClass: Record<AuthorityParty['tier'], string> = {
+  customer: 'text-functional-title',
+  partner: 'text-status-paused',
+  vendor: 'text-functional-text',
+}
+
+export const AuthorityChain = (props: AuthorityChainProps) => {
+  const { parties, expiresIn, onViewChain, className } = props
+
+  if (parties.length === 0) return null
+
+  return (
+    <div
+      className={classNames(
+        'flex flex-wrap items-center gap-x-2 gap-y-1 text-sm',
+        className,
+      )}
+      data-testid="authority-chain"
+      /* The full chain is always in the accessibility tree, whatever the
+         viewport shows. */
+      aria-label={summariseChain(parties)}
+    >
+      <span aria-hidden className="shrink-0">
+        🔗
+      </span>
+
+      {/* Wide viewports: every party, in order. */}
+      <span className="hidden items-center gap-x-2 md:inline-flex">
+        {parties.map((party, i) => (
+          <span
+            key={`${party.tier}-${party.name}`}
+            className="inline-flex items-center gap-x-2"
+          >
+            {i > 0 && (
+              <span aria-hidden className="text-functional-text-light">
+                ←
+              </span>
+            )}
+            <span
+              className={classNames('whitespace-nowrap', tierClass[party.tier])}
+            >
+              {formatParty(party)}
+            </span>
+          </span>
+        ))}
+      </span>
+
+      {/* Narrow viewports: the summary, which still names the hop. */}
+      <span className="inline md:hidden" data-testid="authority-chain-summary">
+        {summariseChain(parties)}
+      </span>
+
+      {expiresIn && (
+        <span className="shrink-0 text-functional-text-light">
+          · support access · expires in {expiresIn}
+        </span>
+      )}
+
+      {onViewChain && (
+        <button
+          type="button"
+          onClick={onViewChain}
+          className="shrink-0 underline underline-offset-2"
+        >
+          view chain
+        </button>
+      )}
+    </div>
+  )
+}
--- a/packages/cube-advisor-ui/src/index.ts
+++ b/packages/cube-advisor-ui/src/index.ts
@@ -1,0 +1,7 @@
+export { AuthorityChain } from './components/AuthorityChain/AuthorityChain'
+export {
+  formatParty,
+  summariseChain,
+  type AuthorityChainProps,
+  type AuthorityParty,
+} from './components/AuthorityChain/formatChain'
--- a/packages/cube-advisor-ui/tsconfig.json
+++ b/packages/cube-advisor-ui/tsconfig.json
@@ -1,0 +1,19 @@
+{
+  "compilerOptions": {
+    "target": "ES2022",
+    "lib": ["ES2022", "DOM", "DOM.Iterable"],
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "jsx": "react-jsx",
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noFallthroughCasesInSwitch": true,
+    "noEmit": true,
+    "skipLibCheck": true,
+    "isolatedModules": true,
+    "resolveJsonModule": true,
+    "types": ["vitest/globals"]
+  },
+  "include": ["src"]
+}
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -93,7 +93,11 @@ export default tseslint.config(
         {
           patterns: [
             {
-              group: ['../**/cube-frontend*', 'packages/**/*'],
+              group: [
+                '../**/cube-frontend*',
+                '../**/cube-advisor*',
+                'packages/**/*',
+              ],
               message:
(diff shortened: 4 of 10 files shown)


packages/cube-advisor-ui/src/components/AuthorityChain/AuthorityChain.tsx
```diff
@@ -0,0 +1,92 @@
...
+ * reader most needs to notice, so it takes the one accent here; the other two
+ * stay in the ordinary text ramp so the chain reads as one sentence.
+ */
+const tierClass: Record<AuthorityParty['tier'], string> = {
+  customer: 'text-functional-title',
+  partner: 'text-status-paused',
+  vendor: 'text-functional-text',
+}
+
+export const AuthorityChain = (props: AuthorityChainProps) => {
```
steven-chiu-bigstack, line 29: Just wanted to confirm if this is the expected appearance of the component (the text may be different based on the provided props):

Image

packages/cube-advisor-ui/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "noUnusedParameters": true,
+    "noFallthroughCasesInSwitch": true,
+    "noEmit": true,
+    "skipLibCheck": true,
+    "isolatedModules": true,
+    "resolveJsonModule": true,
+    "types": ["vitest/globals"]
+  },
+  "include": ["src"]
+}
```
steven-chiu-bigstack, line 19: It is recommended to extend the base `tsconfig.json` to keep our TypeScript setups consistent. Here's the updated content:

```json
{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
    "target": "ES2022",
    "lib": ["ES2022", "DOM", "DOM.Iterable"],
    "jsx": "react-jsx",
    "resolveJsonModule": true,
    "types": ["vitest/globals"]
  },
  "include": ["src"]
}
```

packages/cube-advisor-ui/src/index.ts
steven-chiu-bigstack, line 1: Is the `cube-advisor-ui` package intended to be another UI library? If so, I think we can rename the folder to `packages/cube-frontend-advisor-ui-library` for better clarity and consistency. Be sure to also rename package name in `package.json` in that case.

I can't remember why we chose `cube-frontend-*` as a prefix back then, though. 🤔

packages/cube-advisor-ui/src/components/AuthorityChain/AuthorityChain.tsx
```diff
@@ -0,0 +1,92 @@
...
+    >
+      <span aria-hidden className="shrink-0">
+        🔗
+      </span>
+
+      {/* Wide viewports: every party, in order. */}
+      <span className="hidden items-center gap-x-2 md:inline-flex">
+        {parties.map((party, i) => (
+          <span
+            key={`${party.tier}-${party.name}`}
```
steven-chiu-bigstack, line 53: This is fine if we'll never have parties with the same name and tier. Otherwise, we should also include `role` in the key.

packages/cube-advisor-ui/src/components/AuthorityChain/AuthorityChain.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  if (parties.length === 0) return null
+
+  return (
+    <div
+      className={classNames(
+        'flex flex-wrap items-center gap-x-2 gap-y-1 text-sm',
+        className,
+      )}
+      data-testid="authority-chain"
```
steven-chiu-bigstack, line 40: Note: This is okay for now, but we haven't finalized a `testid` strategy yet, as the QA team may introduce a specific methodology for UI testing in the future.

steven-chiu-bigstack (approved):
Looks good overall! Feel free to merge it when you're ready. 🙏

traviswu-bigstack:

PR #856, opened 2026-08-19
feat(advisor-ui): the global scope switcher and breadcrumb

traviswu-bigstack:

The second cross-cutting element from the redraw brief, present on every screen: **which tier am I acting in, and whose tenant am I inside**.

Stacked on `feat/advisor-ui` (#855), which is awaiting review — retarget to `develop` once that lands.

Refs #854

### Cross-partner selection is structurally impossible, not validated

The brief is blunt about this one:

> Make cross-partner selection structurally impossible in the picker rather than validated after the fact — partners are frequently competitors and this is the screen where a slip is unrecoverable.

So `scopeOptions` **derives** the menu from the viewer's own subtree rather than filtering a global list. The difference is not stylistic: a filter is a check someone can forget to apply, or apply to the wrong list. Deriving means a sibling partner is never in the data the picker renders from at all.

### The prefix bug this is built to not have

Containment is compared **segment by segment**, never as a raw string prefix. `bigstack.north` is a string prefix of `bigstack.northwind` but is not its ancestor — `startsWith` alone hands one partner another partner's entire portfolio. The test fixture contains a partner named `North` specifically to trigger it.

### An end customer gets a label, not an empty dropdown

And deliberately not a *disabled* control either. An empty or greyed picker implies there is somewhere to switch to and that the viewer lacks permission — a different and more alarming claim than "you are here". Nothing is being withheld, so nothing should look withheld.

### The breadcrumb starts at the viewer, not the root

A partner acting inside a customer reads `NorthWind › Acme Corp`, not a chain beginning with Bigstack that implies they are acting under vendor authority. Who acts under whose authority is the **authority chain's** job (#855); conflating the two is how a viewer misreads their own permissions.

## Mutation testing

Seven mutations, **six caught**: naive `startsWith`, an empty viewer scope matching everything, options leaking upward so a partner sees the vendor, the customer getting an empty dropdown, the breadcrumb starting at the root, and unsorted options.

**The seventh survived, and it is not a test gap.** Removing the early return for an out-of-subtree selection changes nothing observable, because the per-segment check inside the loop already yields an empty trail. It is defence in depth, and I kept it because it states the intent at the boundary — but I am not claiming a kill I did not get.

## Verification

`tsc` clean · `eslint` clean · 24 tests pass · prettier applied.

### DIFF
--- a/packages/cube-advisor-ui/src/components/ScopeSwitcher/ScopeSwitcher.tsx
+++ b/packages/cube-advisor-ui/src/components/ScopeSwitcher/ScopeSwitcher.tsx
@@ -1,0 +1,99 @@
+import classNames from 'classnames'
+
+import {
+  breadcrumbTo,
+  isSwitchable,
+  scopeOptions,
+  type TenantNode,
+} from './scope'
+
+export type ScopeSwitcherProps = {
+  /** The viewer's own tenant path — the top of everything they may reach. */
+  viewerPath: string
+  /** The tenant currently being acted in. */
+  selectedPath: string
+  /** Tenants known to the client. Anything outside the viewer's subtree is ignored. */
+  tenants: TenantNode[]
+  onSelect?: (path: string) => void
+  className?: string
+}
+
+const tierLabel: Record<TenantNode['tier'], string> = {
+  vendor: 'Bigstack',
+  partner: 'Partner',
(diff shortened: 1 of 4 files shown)


traviswu-bigstack:

PR #857, opened 2026-08-19
feat(advisor-ui): the provider badge, and the pinned-control affordance

traviswu-bigstack:

The third and last cross-cutting element from the redraw brief: **which brain is active, and who pays for it** — plus the pinned-control affordance it depends on.

Stacked on `feat/advisor-scope-switcher` (#856), itself stacked on #855. Retarget down the chain as each lands.

Refs #854

### The mode is not decoration

Before ADR 0007 the badge answered "which model". With a pooled credential, **"which model" alone no longer tells an operator where their cluster's data went** — so `badgeText` always emits both, and there is a test that it never renders a model name on its own.

```
Claude Sonnet 4.5 · Bigstack-hosted
Claude Opus 4 · your key
gpt-oss-120b · local pack
```

### Cloud egress resolves most-restrictive-wins, and names who decided

ADR 0007 §6: a partner **cannot** loosen `cloud allowed: no` for their customer. So a deny anywhere in the chain beats an allow anywhere, regardless of depth — the mutation for "nearest tier wins" is one of the seven caught.

The resolved policy also carries **which tenant decided**, because the brief requires the reason line to name the tier:

> Disabled by Acme Corp's data-egress policy

Merely saying "disabled" would leave a partner's engineer unable to tell whether they or their customer set it.

### Pinned, not silently absent

The brief allows "absent or pinned with a reason". Pinned is the honest option: **a silently shorter list reads as the product lacking the model**, not as a policy excluding it, and an operator mid-incident shouldn't have to guess which. `PinnedControl` generalises the wave-1 affordance — visibly pinned with a one-line reason, never hidden and never greyed with no explanation.

### Two smaller decisions

- **Gateway degradation renders visibly.** A silent fallback to the local pack means an operator believes they are talking to a model they are not.
- **Data ownership and inference ownership stay separate.** Nothing here derives a mode from a tier — a partner-managed customer might use their own key, and a direct customer might use the pooled one. The brief warns specifically against conflating them.

## Mutation testing

Seven, each caught: nearest-tier-wins, a deny losing to an earlier allow, cloud models dropped rather than pinned, a reason omitting the deciding tier, the local pack pinned too, the badge dropping the mode, and `own-key` misclassified as staying on-cluster.

## Verification

`tsc` clean · `eslint` clean · 35 tests pass · prettier applied.

## What this completes

The three cross-cutting elements the redraw says appear on every screen — authority chain (#855), scope switcher (#856), provider badge (this). The twelve screens come next, translated onto real `Cos*` components rather than ported from the Figma export.

### DIFF
--- a/packages/cube-advisor-ui/src/components/ProviderBadge/ProviderBadge.tsx
+++ b/packages/cube-advisor-ui/src/components/ProviderBadge/ProviderBadge.tsx
@@ -1,0 +1,130 @@
+import classNames from 'classnames'
+
+import {
+  badgeText,
+  isCloudMode,
+  modeLabel,
+  type Model,
+  type ModelAvailability,
+} from './provider'
+
+export type ProviderBadgeProps = {
+  model: Model
+  /** Shown when pooled inference has degraded to the local pack. */
+  degraded?: boolean
+  className?: string
+}
+
+const modeClass: Record<Model['mode'], string> = {
+  pooled: 'text-status-neutral',
+  'own-key': 'text-functional-text',
+  'local-pack': 'text-status-positive-text',
+}
+
+/**
+ * Which brain is active, and who pays for it.
+ *
+ * The mode is not decoration. Before ADR 0007 the badge answered "which model";
+ * with a pooled credential, "which model" alone no longer tells an operator
+ * where their cluster's data went.
+ */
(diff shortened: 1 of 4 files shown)


traviswu-bigstack:

PR #859, opened 2026-08-19
feat(advisor-ui): fleet home, with access and visibility kept apart

traviswu-bigstack:

The first of the twelve screens: **fleet home**, the SaaS landing — enrolled clusters, grouped by tier.

Stacked on `feat/advisor-provider-badge` (#857) → #856 → #855. Retarget down the chain as each lands.

Refs #854

### The failure mode this screen is built around

The brief names it outright:

> each cluster shows whether the viewer currently *has* access or merely visibility. **Those are different states and conflating them is the main way this screen goes wrong.**

So:

- `access` and `visibility` are **separate fields**, never one boolean.
- `fleetCounts` reports **two numbers**, never a total. "12 clusters" tells a partner engineer they can act on twelve when they can act on three.
- The rows differ **structurally**: an accessible cluster is a `<button>`; a visible-only one is text carrying `visible only — no access granted`. Not a dimmed variant of the same thing, because dimming reads as "temporarily unavailable" rather than "you were never granted this".

### Grouping drops, it does not hide

Same rule as the scope switcher: clusters outside the viewer's subtree are removed in `groupFleet`, before the renderer sees them. A sibling partner's clusters never reach the component that draws rows.

A vendor viewer gets `partner → customer` nesting. A partner viewer **does not** get their own name repeated above every customer — they are already inside their own subtree, so it would be a level of noise that says nothing. Direct customers sit in an unparented bucket tagged `direct`; partner-managed ones are tagged `via NorthWind`.

### Smaller calls

- **Staleness is stated, not computed by the reader** — `packs stale` rather than a date to subtract from today.
- **The threshold is exclusive**: exactly 30 days is not yet stale. Off-by-one here would mark a freshly-published pack stale on a boundary day, and there's a mutation for it.

## Mutation testing

Seven, each caught. The two that matter most: **counting visibility as access**, and **letting clusters outside the viewer's subtree into the list**. Also caught: every customer tagged `direct`, a partner viewer nested under their own name, an inclusive staleness threshold, and unstable ordering.

## Verification

`tsc` clean · `eslint` clean · 44 tests pass · prettier applied.

## Scope — please read

This is built from the **written brief**, not pixel-matched to `02 Cube AI Advisor Fleet Home.dc.html`. The data model and the invariants are what I could verify and test; the visual pass against the Claude Design export is still outstanding.

That is deliberate rather than a shortcut: the mockup's component bundle is a *Figma export* (128 components named from layer names), not this library, so the screens have to be translated onto real `Cos*` components. Getting the semantics and the invariants right first means the visual pass is a styling exercise rather than a rewrite. Worth a look from someone with the mockup open.

### DIFF
--- a/packages/cube-advisor-ui/src/components/FleetHome/FleetHome.tsx
+++ b/packages/cube-advisor-ui/src/components/FleetHome/FleetHome.tsx
@@ -1,0 +1,179 @@
+import classNames from 'classnames'
+
+import {
+  fleetCounts,
+  groupFleet,
+  isPackStale,
+  type ClusterRow,
+  type CustomerGroup,
+} from './fleet'
+import type { TenantNode } from '../ScopeSwitcher/scope'
+
+export type FleetHomeProps = {
+  viewerPath: string
+  tenants: TenantNode[]
+  clusters: ClusterRow[]
+  onOpenCluster?: (id: string) => void
+  className?: string
+}
+
+const healthClass: Record<ClusterRow['health'], string> = {
+  ok: 'text-status-positive-text',
+  warning: 'text-status-warning',
+  critical: 'text-status-negative',
+  unknown: 'text-functional-text-light',
+}
+
+const ChannelTag = (props: { channel: CustomerGroup['channel'] }) =>
+  props.channel.kind === 'direct' ? (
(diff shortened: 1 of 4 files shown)


traviswu-bigstack:

PR #853, opened 2026-08-18
fix(nodes): 823 mark an unreadable GPU number Unavailable and split an instance's history link

raven-pan:
#### What type of PR is this?

/kind bug

##### TLDR

Two GPU contract changes land in the node details page:

| # | Contract change | What the UI now does |
| --- | --- | --- |
| 1 | `/gpuCards` answers `null` instead of `0` for a number the host cannot read | Every affected cell says **Unavailable** in grey, with an info icon that names the hardware reason on hover |
| 2 | An attached instance's `links.grafana` becomes `links.workloadHistory` + `links.vramHistory` | Each instance row carries two history links instead of one, each live or disabled with its reason |

---

### Part 1 — a number the host cannot read says "Unavailable", not `null%`

`/gpuCards` used to answer `0` for a number the host cannot read, so "idle" and
"cannot be measured" looked the same. bigstack-oss/cube-cos-api#635 makes those
fields `null` instead. Rendered as-is that reads as a page that failed to load —
two cells printed `null%`, and an attached instance's Grafana link would have
pointed at the wrong VM.

Every affected cell now says **Unavailable** in grey, next to an info icon that
names the hardware reason on hover:

| Card type | Reason on the info icon |
| --- | --- |
| `pgpu` | A passed-through GPU is bound to vfio-pci, so the host cannot read this value. |
| `migBackedVgpu` | NVIDIA reports no device-level utilization once MIG is enabled. |
| anything else | The host reports no value for this card. |

A word beats a dash here. In a column of numbers a `-` reads as zero, or as a cell
still loading; **Unavailable** says the number does not exist, and the info icon
tells the reader where to hover for the reason.

`getUnmeasurableReasonKey` maps the card's resource type to that sentence. The API
deliberately applies **no** per-resource-type rule — a nil source stays nil — so
the card's type is what explains the gap, and the message stops being reachable
once DCGM supplies the numbers.

Sites covered:

- VRAM allocation, VRAM utilization, GPU utilization on the card row
- utilization and memory usage on an attached instance — inline table **and**
  full-view modal
- an instance's history links, which are `null` for a passthrough card: they
  render **disabled with their own reason** rather than vanishing, so the
  operator knows the row is complete

##### What reads Unavailable, per resource type

On the GPU card row:

| Resource type | VRAM Allocation | VRAM Utilization | GPU Utilization | Why |
| --- | --- | --- | --- | --- |
| `unset` | number | number | number | the host owns the whole card |
[...]

PR #860, opened 2026-08-19
fix(nodes): 1293 block Edit GPU Type while a card is still changing

### LINKED ISSUE #1293: [Bug]: Edit GPU Type stays clickable while the same card is still processing a change
## Describe the bug

`Edit GPU Type` in the GPU card row menu stays clickable while a previous change to
that same card is still running. The card already tells the UI it is busy — the row
shows a spinner next to the card name, driven by `status.isProcessing` — but the
menu entry ignores it and only checks `status.current`. An operator can therefore
open the edit modal and submit a second resource-type change before the first one
settles.

The same node page already gets this right for block devices: their row actions are
replaced by a spinner while the device or its OSD is processing. GPU rows only got
the spinner, not the gating.

## Affected Component / Service

`cube-cos-ui` — `Nodes > <node> > GPU Resources`.

## Version

- Product/Fixpack version: CubeCOS 3.2 (GPU productization), reproduced on a 3.1.10 appliance build with the current `develop` UI
- Component version: `cube-cos-ui` `develop`

## To Reproduce

1. Go to `Nodes > <node>` on a node that has a GPU card whose status is `Idle` or `Unassigned`.
2. Open the `⋯` menu on that card's row and click `Edit GPU Type`.
3. Pick a different resource type and confirm, so the card starts changing. A spinner appears next to the card name.
4. While that spinner is still showing, open the `⋯` menu on the same row again.
5. `Edit GPU Type` is still enabled, and clicking it opens the modal for a second change.

## Expected behavior

While a card reports `status.isProcessing`, `Edit GPU Type` is disabled, so the
operator cannot queue a second change on top of an unfinished one. The card's
existing spinner stays as the explanation for why the entry is unavailable.

## Root Cause (if known)

The disable condition covers only the steady-state status, not the in-flight flag:

```tsx
// packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/NodeGpuResources.tsx:208
disabled={row.status.current === GPUCardStatus.InUse}
```

`status` carries both fields (`{ current, isProcessing }` — the API returns an
object precisely so an in-flight operation can be represented), and the same file
already reads the other one for display:

```tsx
// NodeGpuResources.tsx:106
{row.status.isProcessing && <CosLoadingSpinner variant="dot120" />}
```

For comparison, the Devices table on the same page hides its actions behind that
flag:

- `devices/nodeDevicesUtils.ts:38` — `isDeviceOrOSDProcessing()`
- `devices/NodeDevices.tsx:130` and `:161` — a spinner replaces the row actions while processing

So this is a missed condition in the GPU table, not a missing capability.

Unclear and worth checking as part of the fix: whether the API rejects a
concurrent update. `docs.yaml` documents `409` on
`PUT /nodes/{nodeName}/gpuCards/{gpuId}`, but its only example is
`GPU card does not support 'sriovVgpu' resource type`. If the API accepts the
second request, the UI gate is the only thing standing between an operator and two
overlapping resource-type changes on one card.

## Suggested Fix (Optional)

Add the flag to the condition:

```tsx
disabled={
  row.status.current === GPUCardStatus.InUse || row.status.isProcessing
}
```

Two things to settle while fixing it:

- Whether the other two menu entries (`View Workload History`, `View VRAM History`) should also be gated. They only open a Grafana tab, so they are harmless during a change, and the panel is where an operator would watch the change land — argues for leaving them alone.
- Whether the API rejects the second `PUT`. If it does not, the fix belongs on both sides.

## Screenshots (Optional)

N/A — the spinner and the enabled menu entry are both visible on the row described
in the steps.

## Additional Context (Optional)

- Cluster topology: any node with at least one GPU card; a single control-converged node is enough
- Hardware: any card that supports more than one resource type, so the type can actually be changed
- Found while reviewing the GPU row menu for cubecos#824 / cubecos#825, which add the two history entries next to `Edit GPU Type`. Those PRs do not change this condition.

## QA Verification

| # | Command / steps | Expected outcome |
|---|---|---|
| 1 | On a node with an `Idle` GPU card, change its resource type from the `⋯` menu and confirm | A spinner appears next to the card name while the change runs |
| 2 | While that spinner is showing, reopen the same row's `⋯` menu | `Edit GPU Type` is greyed out and does not open the modal |
| 3 | Wait for the spinner to disappear, then reopen the `⋯` menu | `Edit GPU Type` is clickable again |
| 4 | Repeat step 2 on a card whose status is `In-use` (no change running) | `Edit GPU Type` stays greyed out, as it does today |
| 5 | Repeat step 2 on a card whose status is `Unassigned` with no change running | `Edit GPU Type` is clickable |

- [ ] **QA Status declared at Done** — when this issue reaches `Done`, set the board's
      `QA Status` field to `Not needed` or `Ready to QA`. Never leave it empty on a Done
      ticket; `In QA` / `Verified` are QA's own transitions.

## Output artifacts (Definition of Done)

- [ ] **Handbook knowledge update** — land the durable, team-readable knowledge from this
      work into the bigstack-handbook cubecos kb (`kb/cubecos/…`) via
      `/bigstack-core:save-to-handbook` (Topic / Runbook / Known-issue / ADR as fits).

raven-pan:
#### What type of PR is this?

/kind bug

`Edit GPU Type` stayed clickable while the same card was still applying a previous
change. The row already showed a spinner for that state — the API reports
`status.isProcessing` next to `status.current` precisely so an in-flight operation
can be represented — but the menu entry only read `status.current`. An operator
could open the modal again and submit a second resource-type change on top of an
unfinished one.

Changing a resource type re-partitions the card, so the entry now needs the card to
be **free and settled**:

| `status.current` | `status.isProcessing` | Edit GPU Type | Change |
| --- | --- | --- | --- |
| `inUse` | `false` | disabled | same as before |
| `inUse` | `true` | disabled | same as before |
| `idle` | `true` | **disabled** | **this PR** |
| `unassigned` | `true` | **disabled** | **this PR** |
| `idle` | `false` | enabled | same as before |
| `unassigned` | `false` | enabled | same as before |

The two history entries stay clickable. They open a Grafana panel, which is where
an operator would watch the change land.

##### Screenshots

Same card in all three: `NVIDIA A100 40GB`, status `Idle`, a change still running —
the spinner next to the card name is the same signal the fix now reads.

**Before — the entry is clickable during the change**

1293-0-before

**After — the entry is greyed out**

1293-1-idle-processing

**After — an idle card with nothing running still opens the modal**

1293-2-idle-settled

#### QA perspective

**Where.** `Nodes > <node> > GPU Resources`, the `⋯` menu on a GPU card row.

**What changed.** Only when a card is mid-change. A card that is idle, unassigned,
or in use behaves exactly as before. No API contract changes, so no other repo is
involved.

**Check list**

- [ ] On a card whose status is `Idle`, change its resource type from the `⋯` menu and confirm. A spinner appears next to the card name.
- [ ] While that spinner shows, reopen the same row's `⋯` menu: `Edit GPU Type` is greyed out and clicking it opens nothing.
- [ ] After the spinner disappears, reopen the menu: `Edit GPU Type` is clickable again.
- [ ] On a card with status `In-use` and no change running: `Edit GPU Type` stays greyed out, as before this PR.
- [ ] On a card with status `Unassigned` and no change running: `Edit GPU Type` is clickable.
- [ ] While a card is mid-change, `View Workload History` and `View VRAM History` on that row stay clickable and still open Grafana.

#### Developer perspective

Two files under `src/pages/node/[name]/_components/NodeGpuResources/`:

- `utils.ts` — new `isGpuTypeEditDisabled(status)` holding both conditions, next to the other row predicates.
- `NodeGpuResources.tsx` — the menu entry calls it instead of comparing `status.current` inline.

`utils.test.ts` covers all six status combinations from the table above.

Verified in a browser against the MSW GPU mocks, which already carry three cards
with `isProcessing: true`. Every one of the 12 rows was read back out of the DOM
and matched what the table predicts — including `0000:07:00.0`, which is `In-use`
*and* processing, and the three settled `Idle` rows that must stay clickable.
Console clean. The "before" screenshot comes from reverting just that one line with
the same mocks, so the two shots differ only by the fix.

`pnpm lint` passes; `pnpm test` reports 104 passing tests, 3 of them new.

Fixes https://github.com/bigstack-oss/cubecos/issues/1293

> Note

- **Open question from the ticket, not settled here:** whether the API rejects a
  concurrent `PUT /nodes/{nodeName}/gpuCards/{gpuId}`. `docs.yaml` documents `409`
  for that endpoint, but its only example is
  `GPU card does not support 'sriovVgpu' resource type`. If the API accepts a
  second request, this UI gate is the only thing preventing two overlapping
  changes, and the backend should get the same guard.
- The same node page already gates block-device actions this way
  (`devices/nodeDevicesUtils.ts:38`, `devices/NodeDevices.tsx:130`), which is where
  the pattern comes from. GPU rows had the spinner but not the gating.
- Touches the same menu block as #852 and #853, but a different line, so it merges
  either way round.

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/NodeGpuResources.tsx
+++ b/packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/NodeGpuResources.tsx
@@ -25,7 +25,11 @@ import { nodesApi } from '@cube-frontend/web-app/api/cosApi'
 import { usePolling } from '@cube-frontend/web-app/hooks/usePolling'
 import { GpuDetailsFullViewModal } from './GpuDetailsFullViewModal/GpuDetailsFullViewModal'
 import { EditGPUResourceModal } from './EditGPUResourceModal/EditGPUResourceModal'
-import { GpuResourceRow, GpuTypeLabelKeyMap } from './utils'
+import {
+  GpuResourceRow,
+  GpuTypeLabelKeyMap,
+  isGpuTypeEditDisabled,
+} from './utils'
 import { useTranslation } from 'react-i18next'
 import { ParseKeys } from 'i18next'
 
@@ -177,7 +181,7 @@ const NodeGpuResources = (props: NodeGpuResourcesProps) => {
           title={t('nodes.details.editGpuType')}
           type="plain"
           onClick={() => openResourceEditModal(row)}
-          disabled={row.status.current === GPUCardStatus.InUse}
+          disabled={isGpuTypeEditDisabled(row.status)}
         />
       </CosOverflowMenu>
     )
--- a/packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/utils.test.ts
+++ b/packages/cube-frontend-web-app/src/pages/node/[name]/_components/NodeGpuResources/utils.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test } from 'vitest'
-import { GPUResourceType } from '@cube-frontend/api'
-import { isProfileRemainingSupported } from './utils'
+import { GPUCardStatus, GPUResourceType } from '@cube-frontend/api'
+import { isGpuTypeEditDisabled, isProfileRemainingSupported } from './utils'
 
 describe('isProfileRemainingSupported', () => {
   test('returns true for MIG-backed vGPU, which limits instances per profile', () => {
@@ -16,5 +16,46 @@ describe('isProfileRemainingSupported', () => {
   test('returns false for resource types without profiles', () => {
     expect(isProfileRemainingSupported(GPUResourceType.Pgpu)).toBe(false)
     expect(isProfileRemainingSupported(GPUResourceType.Unset)).toBe(false)
+  })
+})
+
+describe('isGpuTypeEditDisabled', () => {
+  test('blocks a card a VM already holds', () => {
+    expect(
+      isGpuTypeEditDisabled({
+        current: GPUCardStatus.InUse,
+        isProcessing: false,
+      }),
+    ).toBe(true)
+  })
+
+  test('blocks a card whose previous change has not settled', () => {
+    expect(
+      isGpuTypeEditDisabled({
+        current: GPUCardStatus.Idle,
+        isProcessing: true,
+      }),
+    ).toBe(true)
+    expect(
+      isGpuTypeEditDisabled({
+        current: GPUCardStatus.Unassigned,
+        isProcessing: true,
+      }),
+    ).toBe(true)
+  })
+
+  test('allows an idle or unassigned card that is not changing', () => {
+    expect(
+      isGpuTypeEditDisabled({
+        current: GPUCardStatus.Idle,
+        isProcessing: false,
+      }),
+    ).toBe(false)
+    expect(
+      isGpuTypeEditDisabled({
+        current: GPUCardStatus.Unassigned,
+        isProcessing: false,
+      }),
+    ).toBe(false)
   })
 })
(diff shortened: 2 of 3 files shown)


steven-chiu-bigstack (commented):
LGTM!

steven-chiu-bigstack (approved):

PR #861, opened 2026-08-19
feat(advisor-ui): cluster workspace — the grounding vocabulary

traviswu-bigstack:

The main screen's chat surface: **citation chips with provenance, the coverage meter, ungrounded segments, conflict cards and command validation**. The layout itself is unchanged from the base brief — the partner-tier brief says only these elements change.

Stacked on #859 → #857 → #856 → #855.

Refs #854

### Audience filtering is the load-bearing rule, and it lives in the model

`coverageOf` counts only entries the viewer may see — **in both numerator and denominator**. So the same question legitimately shows different coverage to different roles. That is correct, not a bug:

> reporting 5/7 to a customer whose two unseen entries are internal would disclose that two internal entries exist

The unread list is filtered the same way, because the brief is explicit that internal titles must not reach customer roles *"even in 'not consulted' lists"*. Filtering in the renderer would leave the counts wrong even when the list looked right.

### Provenance at a glance, not on hover

`kb: ceph-slow-ops · Bigstack` versus `kb: nw-storage-runbook · Partner: NorthWind` — the layer is in the chip **text** and the colour, per the brief. The coverage meter splits by layer (`Based on 2/3 · 1 Bigstack · 1 Partner`), and a layer that contributed nothing is omitted rather than printed as a zero.

### Conflicts have no winner, and same-layer disagreements aren't conflicts

`Conflict` has **no winner field**, so the type cannot express one — the brief requires both sides shown with provenance and no winner picked.

`isLayerConflict` additionally requires **two distinct layers**. Two Bigstack entries disagreeing is a knowledge-base inconsistency for us to fix, not something to hand an operator mid-incident to adjudicate.

### Absence of a marker is itself a claim

- An **uncited segment** renders visibly differently (amber, left rule, `(ungrounded)`), because it must not look like a grounded one.
- A **command block always carries a badge**. A block with no badge reads as endorsed, so unsupported is stated rather than omitted.

## Mutation testing — and one test that was passing for the wrong reason

Nine mutations, **eight caught first time**. Worth reporting the ninth honestly:

**"tool citations counted as knowledge coverage" survived.** My test used a tool ref of `cluster check @14:02`, which matches no outline slug — so removing the `kind === 'kb'` filter changed nothing and the test passed regardless of whether the filtering existed. I rewrote it so the tool ref **collides with a real entry slug** (`ceph-recovery`), which is also the realistic case. It now fails against the mutation.

The other eight: internal entries visible to customers, the denominator counting unseeable entries, internal titles leaking into the unread list, the layer dropped from a chip, a partner labelled as Bigstack, same-layer disagreement treated as a conflict, an uncited segment reported as grounded, and an unsupported command getting no badge.

## Verification

`tsc` clean · `eslint` clean · 61 tests pass · prettier applied.

## Scope

Built from the **written brief**, not pixel-matched to `03 Cube AI Advisor Cluster Workspace.dc.html`. Semantics and invariants first; the visual pass against the Design export is outstanding, and needs someone with the mockup open.

Not included from this screen: the **tool-activity trace**, **approval cards** for bounded probes, and the **right-hand panel** that swaps between source viewer / tool output / session history. Those are stateful and belong with the API wiring rather than with the grounding vocabulary.

### DIFF
--- a/packages/cube-advisor-ui/src/components/ClusterWorkspace/ClusterWorkspace.tsx
+++ b/packages/cube-advisor-ui/src/components/ClusterWorkspace/ClusterWorkspace.tsx
@@ -1,0 +1,162 @@
+import classNames from 'classnames'
+
+import {
+  citationLabel,
+  commandBadge,
+  coverageLabel,
+  isGrounded,
+  isLayerConflict,
+  type AnswerSegment,
+  type Citation,
+  type CommandBlock,
+  type Conflict,
+  type Coverage,
+} from './answer'
+
+const layerClass: Record<Citation['layer'], string> = {
+  bigstack: 'text-status-neutral',
+  partner: 'text-status-paused',
+}
+
+/**
+ * A citation chip.
+ *
+ * The layer is in the text and in the colour, because the brief requires
+ * provenance to be distinguishable at a glance rather than on hover.
+ */
+export const CitationChip = (props: {
+  citation: Citation
+  onOpen?: (ref: string) => void
+}) => {
+  const { citation, onOpen } = props
+  return (
+    <button
+      type="button"
+      onClick={() => onOpen?.(citation.ref)}
+      className={classNames(
+        'rounded border border-functional-border-divider px-1 text-xs',
+        layerClass[citation.layer],
+      )}
+      data-testid="citation-chip"
+      data-layer={citation.layer}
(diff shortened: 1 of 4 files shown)


traviswu-bigstack:

PR #863, opened 2026-08-20
fix(nodes): 1305 poll GPU cards every 30s, and not at all in a hidden tab

raven-pan:
#### What type of PR is this?

/kind bug

##### TLDR

The GPU Resources page polled a 9.5-second endpoint every 5 seconds, including while nobody was looking at the tab. It now polls every 30 s, and not at all in a hidden tab.

---

### The measurement

`/gpuCards` on `cn13`, 8 sequential runs, 4 cards, 35.9 KB response:

| What | Value |
| --- | --- |
| `/gpuCards` total time | **9.00 – 9.52 s** |
| Same node's `/nodes/cn13`, for comparison | **0.12 s** |
| 3 requests in parallel | 10.8 – 11.0 s each — the work does not serialize |
| Poll interval before this PR | 5 s |
| Real refresh cycle before this PR | **~14.5 s** — `usePolling` starts its delay *after* the response |
| Server load, one tab open | ~4.2 requests/min × 9.5 s ≈ **65 % of one handler busy, continuously** |

The endpoint's own latency is the real defect and stays open in bigstack-oss/cubecos#1305. This PR fixes the two things that are the UI's to fix.

### What changed

**1. The interval is 30 s.** A 5 s interval described a refresh rate the endpoint can never deliver — the constant said 5 s while the page actually refreshed every ~14.5 s. The values on this table move when an operator carves a card or attaches a VM, not second by second; live utilization belongs in the Grafana history each row already links to.

**2. Polling stops in a hidden tab.** `usePolling` gains an opt-in `pauseWhenHidden`, which stops on `visibilitychange` to hidden and polls once as soon as the tab is visible again. **Only the GPU page opts in** — the other 13 polling call sites keep their current behaviour.

**3. The visibility check also runs inside the polling callback**, not just around the timer. A request already in flight when the tab hides re-schedules the next run from its own `finally` block, so clearing the timer alone lets polling quietly resume while hidden. There is a test for exactly that.

**4. `useSequentialInterval`'s `startInterval` takes an `{ immediate }` override**, so resuming on return fetches at once instead of waiting out a full interval.

##### Repo dependency chain

| Repo | Role |
| --- | --- |
| `cube-cos-api` | owns the 9.5 s — the real fix lives here, not merged, tracked in bigstack-oss/cubecos#1305 |
| `cube-cos-ui` | this PR — stops the UI from asking more often than the API can answer |

No API or spec change is needed for this PR. It behaves the same against today's deployed `cube-cos-api`.

##### Expectation — check list

Open **Nodes → pick a node with GPU cards → GPU Resources**, with DevTools → Network filtered to `gpuCards`.

- [ ] One `gpuCards` request every ~30 s while the tab is in the foreground — not every 5 s
- [ ] Switch to a different browser tab for one minute: **no** `gpuCards` request is issued during that minute
- [ ] Switch back: one request fires immediately, and the ~30 s rhythm resumes
- [ ] Minimise the whole browser window for a minute: same as switching tabs — no requests
- [ ] Leave the page (go to another route) and come back: polling restarts, and no duplicate request storm
- [ ] The table's numbers still update — carve a card or attach a VM elsewhere, wait 30 s, and the row reflects it
- [ ] Other pages that poll are unchanged: Home Health, Home Overview, Settings, Maintenance → Tunings still refresh at their own cadence in the background

##### Config note

Nothing to configure. The interval is a constant in the page, not a setting.

---

### Developer

| File | Change |
| --- | --- |
| `hooks/usePolling.ts` | new `pauseWhenHidden` option: a `visibilitychange` listener plus a check inside the polling callback |
| `hooks/useSequentialInterval/useSequentialInterval.ts` | `startInterval` accepts `{ immediate }` to override the hook-level default for one restart |
| `pages/node/[name]/_components/NodeGpuResources/NodeGpuResources.tsx` | `GPU_POLLING_INTERVAL` 5 s → 30 s, with the measurement in a comment; opts into `pauseWhenHidden` |
| `hooks/usePolling.test.ts` | **new.** Six cases |
| `package.json` / `pnpm-lock.yaml` | `jsdom` and `@testing-library/react` as devDependencies |

##### The monorepo's first DOM test

There was no DOM test environment before this PR — all 101 existing tests are pure functions in the node environment. A `visibilitychange` listener cannot be tested without one, so this PR adds `jsdom` and `@testing-library/react`.

`jsdom` is scoped through a `// @vitest-environment jsdom` docblock in the one test file that needs it, **not** set globally in `vite.config.ts`, so the other six test files keep running in the node environment at their current speed.

The six cases:

| Case | Asserts |
| --- | --- |
| polls while visible | the interval keeps firing |
| stops while hidden | zero calls across 10 intervals |
| polls once on return | exactly one catch-up call, not a backlog |
| in-flight request resolves after hiding | the `finally` re-schedule does not resume polling |
| unmount | no calls after unmount |
| no `pauseWhenHidden` | visibility is ignored — the existing call sites are unaffected |

Refs #1305

This PR does **not** close #1305. The 9.5 s endpoint is still there; this only stops the UI from making it worse. Close #1305 with the `cube-cos-api` change.

> Note

- **Written test-first.** The three new-behaviour cases were watched failing before the implementation existed (`expected 1 call, got 11`), and the timing assumption in one of them was wrong at first — `usePolling` defaults to `immediate: false`, so the first poll lands one interval in. The test was corrected, not the production code.
- **The in-flight race is the subtle part.** Clearing the timer on `visibilitychange` is not enough: a request already awaiting re-arms the timer from its own `finally`. That is why the check is duplicated inside the callback. Removing either half breaks a test.
- **`pauseWhenHidden` is opt-in.** Making it the default would change 13 other polling call sites in one PR. If the team wants it everywhere, that is a separate change with its own QA pass.
- **30 s follows an existing precedent** — `CHART_PAGE_POLLING_INTERVAL` is already `30 * 1000`.
- **No screenshots.** Nothing on screen changes; the difference is request cadence, which is visible only in the Network tab.
- **Two pre-existing failures you will hit locally, neither from this PR.** `pnpm lint` fails in `cube-frontend-keycloak-login`: that package's `tsc` script has no `typescript` dependency of its own and only ever resolved through a stale hoisted symlink, so a clean `pnpm install` breaks it. Every other package type-checks clean, `pnpm eslint` is clean, and `pnpm prettier` flags only an untracked local tool cache. Separately, `vitest` prints "something prevents Vite server from exiting" — that happens on an untouched existing test file too.

Verified: `pnpm test` reports **107 passing tests, 6 of them new**. `pnpm eslint` clean. All packages except the pre-existing `keycloak-login` breakage type-check.

#### Additional documentation

```docs

```

### DIFF
--- a/packages/cube-frontend-web-app/package.json
+++ b/packages/cube-frontend-web-app/package.json
@@ -42,11 +42,13 @@
     "zustand": "^5.0.12"
   },
   "devDependencies": {
+    "@testing-library/react": "^16.3.2",
     "@types/react": "catalog:",
     "@types/react-dom": "catalog:",
     "@vitejs/plugin-react-swc": "catalog:",
     "autoprefixer": "catalog:",
     "code-inspector-plugin": "catalog:",
+    "jsdom": "^30.0.1",
     "msw": "^2.7.0",
     "postcss": "catalog:",
     "tailwind-merge": "catalog:",
--- a/packages/cube-frontend-web-app/src/hooks/usePolling.test.ts
+++ b/packages/cube-frontend-web-app/src/hooks/usePolling.test.ts
@@ -1,0 +1,128 @@
+// @vitest-environment jsdom
+import { act, cleanup, renderHook } from '@testing-library/react'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { usePolling } from './usePolling'
+
+const setTabVisibility = (state: 'visible' | 'hidden') => {
+  Object.defineProperty(document, 'visibilityState', {
+    configurable: true,
+    get: () => state,
+  })
+  document.dispatchEvent(new Event('visibilitychange'))
+}
+
+const advance = (ms: number) =>
+  act(async () => {
+    await vi.advanceTimersByTimeAsync(ms)
+  })
+
+describe('usePolling', () => {
+  beforeEach(() => {
+    vi.useFakeTimers()
+    setTabVisibility('visible')
+  })
+
+  afterEach(() => {
+    cleanup()
+    vi.useRealTimers()
+  })
+
+  it('keeps polling while the tab stays visible', async () => {
+    const fetch = vi.fn().mockResolvedValue(undefined)
+
+    renderHook(() => usePolling(fetch, 1000, { pauseWhenHidden: true }))
+
+    // usePolling defaults to `immediate: false`, so the first poll lands one
+    // interval in, not on mount.
+    await advance(1000)
+    expect(fetch).toHaveBeenCalledTimes(1)
+
+    await advance(3000)
+    expect(fetch.mock.calls.length).toBeGreaterThan(1)
+  })
+
+  it('stops polling while the tab is hidden', async () => {
+    const fetch = vi.fn().mockResolvedValue(undefined)
+
+    renderHook(() => usePolling(fetch, 1000, { pauseWhenHidden: true }))
+
+    await advance(1000)
+    const callsBeforeHiding = fetch.mock.calls.length
+    expect(callsBeforeHiding).toBe(1)
+
+    act(() => setTabVisibility('hidden'))
+    await advance(10_000)
+
+    expect(fetch).toHaveBeenCalledTimes(callsBeforeHiding)
+  })
+
+  it('polls once as soon as the tab becomes visible again', async () => {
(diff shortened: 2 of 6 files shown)


steven-chiu-bigstack (approved):
LGTM, nice work! 🚀

